PR #4, opened 2024-03-01
Add Retry Mechanism for Gradlew, Terraform, and Eksctl

harrryr:
*Issue #, if available:*
There are occasional failures in the canary due to internet transient issue while downloading some external libraries

*Description of changes:*
Adding a retry mechanism to Gradlew, Terraform and Eksctl. Also manually implement Terraform rather than using the setup terraform Github Action

Test Run: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8104615686
Test Run after review: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8115879115
Test Run after changing eq sign: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8117425277
Test Run after refactoring: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8118227976
Test Run after editing curl feature flag: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8118610550

### DIFF
--- a/.github/workflows/appsignals-e2e-ec2-test.yml
+++ b/.github/workflows/appsignals-e2e-ec2-test.yml
@@ -158,10 +158,10 @@ jobs:
       - name: Call all test APIs
         continue-on-error: true
         run: |
-          curl -S -o /dev/null http://${{ env.MAIN_SERVICE_ENDPOINT }}/outgoing-http-call/
-          curl -S -o /dev/null http://${{ env.MAIN_SERVICE_ENDPOINT }}/aws-sdk-call/
-          curl -S -o /dev/null http://${{ env.MAIN_SERVICE_ENDPOINT }}/remote-service?ip=${{ env.REMOTE_SERVICE_IP }}/
-          curl -S -o /dev/null http://${{ env.MAIN_SERVICE_ENDPOINT }}/client-call/
+          curl -S -s http://${{ env.MAIN_SERVICE_ENDPOINT }}/outgoing-http-call/
+          curl -S -s http://${{ env.MAIN_SERVICE_ENDPOINT }}/aws-sdk-call/
+          curl -S -s http://${{ env.MAIN_SERVICE_ENDPOINT }}/remote-service?ip=${{ env.REMOTE_SERVICE_IP }}/
+          curl -S -s http://${{ env.MAIN_SERVICE_ENDPOINT }}/client-call/
           
       - name: Build Gradlew
         run: |
--- a/.github/workflows/appsignals-e2e-ec2-test.yml
+++ b/.github/workflows/appsignals-e2e-ec2-test.yml
@@ -31,6 +31,8 @@ env:
   LOG_GROUP_NAME: /aws/appsignals/generic
   TEST: ${{ inputs.test }}
   GET_ADOT_JAR_COMMAND: "wget -O adot.jar https://github.com/aws-observability/aws-otel-java-instrumentation/releases/latest/download/aws-opentelemetry-agent.jar"
+  TEST_RESOURCES_FOLDER: /home/runner/work/aws-application-signals-test-framework/aws-application-signals-test-framework
+
 
 jobs:
   e2e-ec2-test:
@@ -65,16 +67,21 @@ jobs:
           aws-region: ${{ inputs.aws-region }}
 
       - name: Set up terraform
-        uses: hashicorp/setup-terraform@v3
-        with:
-          terraform_wrapper: false
+        run: |
+          source ${{ env.TEST_RESOURCES_FOLDER }}/.github/workflows/util/execute_and_retry.sh
+          execute_and_retry 2 "wget -O- https://apt.releases.hashicorp.com/gpg | sudo gpg --dearmor -o /usr/share/keyrings/hashicorp-archive-keyring.gpg"   
+          echo "deb [signed-by=/usr/share/keyrings/hashicorp-archive-keyring.gpg] https://apt.releases.hashicorp.com $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/hashicorp.list
+          sudo apt update && sudo apt install terraform
+
+      - name: Initiate Terraform
+        working-directory: terraform/ec2
+        run: |
+          source ${{ env.TEST_RESOURCES_FOLDER }}/.github/workflows/util/execute_and_retry.sh
+          execute_and_retry 2 "terraform init && terraform validate" "rm -rf .terraform && rm -rf .terraform.lock.hcl"
 
       - name: Deploy sample app via terraform and wait for endpoint to come online
         working-directory: terraform/ec2
         run: |
-          terraform init
-          terraform validate
-          
           # Attempt to deploy the sample app on an EC2 instance and wait for its endpoint to come online. 
           # There may be occasional failures due to transitivity issues, so try up to 2 times. 
           # deployment_failed of 0 indicates that both the terraform deployment and the endpoint are running, while 1 indicates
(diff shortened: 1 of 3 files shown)


.github/workflows/appsignals-e2e-eks-test.yml
```diff
@@ -158,15 +165,6 @@ jobs:
...
           
-              # If the workflow provides a specific ADOT image to test, patch the deployment and restart CW agent related pods
-              if [ ${{ inputs.appsignals-adot-image-name }} != "" ]; then
-                kubectl patch deploy -namazon-cloudwatch amazon-cloudwatch-observability-controller-manager --type='json' \
-                -p='[{"op": "replace", "path": "/spec/template/spec/containers/0/args/0", "value": "--auto-instrumentation-java-image=${{ inputs.appsignals-adot-image-name }}"}]'
-          
-                kubectl delete pods --all -n amazon-cloudwatch
-                kubectl wait --for=condition=Ready pod --all -n amazon-cloudwatch
-              fi
-
```
majanjua-amzn, line 169: Why remove?

.github/workflows/appsignals-e2e-ec2-test.yml
```diff
@@ -65,16 +67,21 @@ jobs:
...
+          source ${{ env.TEST_RESOURCES_FOLDER }}/.github/workflows/util/execute_and_retry.sh
+          execute_and_retry 2 "wget -O- https://apt.releases.hashicorp.com/gpg | sudo gpg --dearmor -o /usr/share/keyrings/hashicorp-archive-keyring.gpg"   
+          echo "deb [signed-by=/usr/share/keyrings/hashicorp-archive-keyring.gpg] https://apt.releases.hashicorp.com $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/hashicorp.list
+          sudo apt update && sudo apt install terraform
+
+      - name: Initiate Terraform
+        working-directory: terraform/ec2
+        run: |
+          source ${{ env.TEST_RESOURCES_FOLDER }}/.github/workflows/util/execute_and_retry.sh
+          execute_and_retry 2 "terraform init && terraform validate" "rm -rf .terraform && rm -rf .terraform.lock.hcl"
```
majanjua-amzn, line 80: Are we sure this is more reliable than the `hashicorp/setup-terraform` action? No point in making this change without testing it rigorously. I suggest running the test many times

.github/workflows/util/execute_and_retry.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+execute_and_retry () {
+  retry_counter=0
+  max_retry=$1
+  while [ $retry_counter -lt $max_retry ]; do
+   deployment_failed=0
```
majanjua-amzn, line 7: This is no longer specific to the deployment code, so make it clear this tracks attempt failures.
```suggestion
   attempt_failed=0
```

.github/workflows/util/execute_and_retry.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+execute_and_retry () {
```
majanjua-amzn, line 3: Put a comment to say what this is used for and how to use it. What happens if we use this function to run something like `terraform apply`? If it fails it will retry without running `terraform destroy` first, right?

.github/workflows/appsignals-e2e-eks-test.yml
```diff
@@ -244,10 +242,15 @@ jobs:
...
-          curl -S -s -o /dev/null http://${{ env.APP_ENDPOINT }}/client-call/
+          curl -S -o /dev/null http://${{ env.APP_ENDPOINT }}/outgoing-http-call/
+          curl -S -o /dev/null http://${{ env.APP_ENDPOINT }}/aws-sdk-call/
+          curl -S -o /dev/null http://${{ env.APP_ENDPOINT }}/remote-service?ip=${{ env.REMOTE_SERVICE_POD_IP }}/
+          curl -S -o /dev/null http://${{ env.APP_ENDPOINT }}/client-call/
+
+      - name: Build Gradlew
+        run: |
+          source ${{ env.TEST_RESOURCES_FOLDER }}/.github/workflows/util/execute_and_retry.sh
+          execute_and_retry 2 "./gradlew"
```
majanjua-amzn, line 253: Instead of this we should just build the validator into a JAR file and run it using that. We can host said JAR file in an S3 bucket on the test accounts and push it via another workflow. This way we do not need to worry about build issues.

majanjua-amzn: We can do this in the meantime though if you want to track that as a separate improvement.

.github/workflows/util/execute_and_retry.sh
```diff
@@ -0,0 +1,25 @@
...
+   deployment_failed=0
+   eval "$2" || deployment_failed=$?
+
+   if [ $deployment_failed -eq 1 ]; then
+     eval "$3"
+     retry_counter=$(($retry_counter+1))
+     sleep 5
+   else
+     break
+   fi
```
majanjua-amzn, line 16: Is this saying "If the command $2 doesn't fail, then run $3"? Can you clarify what the inputs to this function are?

.github/workflows/appsignals-e2e-eks-test.yml
harrryr: This part of the code was for release testing on ADOT. It won't be used in this repository anymore

.github/workflows/appsignals-e2e-ec2-test.yml
harrryr: Regardless of whether it is more reliable or not, this error:
`(Set up Terraform) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 close listeners added to [TLSSocket]. Use emitter.setMaxListeners() to increase limit`
is a Node.js issue that we can't fix since `hashicorp/setup-terraform` uses Node.js to perform its action.

.github/workflows/util/execute_and_retry.sh
harrryr: There is a third parameter to perform cleanup action. Will mention in a comment

.github/workflows/appsignals-e2e-eks-test.yml
harrryr: Will do thanks

.github/workflows/util/execute_and_retry.sh
harrryr: $1: Retry count
$2: Command to attempt
$3: Cleanup Command

.github/workflows/appsignals-e2e-ec2-test.yml
harrryr: Also, the commands used are the recommended installation step in Hashicorp's [website](https://developer.hashicorp.com/terraform/tutorials/aws-get-started/install-cli)

Additionally, we use wget to retrieve files in multiple instances during the e2e testing, so it will probably suffer from the same internet transient issue

.github/workflows/appsignals-e2e-eks-test.yml
majanjua-amzn: Doesn't ADOT release testing refer to this repo now though?

.github/workflows/appsignals-e2e-ec2-test.yml
majanjua-amzn: Got it, let's try it

.github/workflows/util/execute_and_retry.sh
majanjua-amzn: Ah makes sense, thanks

.github/workflows/util/execute_and_retry.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+# This function is for retrying commands in the case they fail. It accepts three arguments
+# $1: Number of retries it will attempt
+# $2: Command to execute
+# $3: Clean up commands
```
majanjua-amzn, line 6: Should note that 3 is optional, but expected if 2 does something that needs to be cleaned if it fails.

.github/workflows/util/execute_and_retry.sh
```diff
@@ -0,0 +1,29 @@
...
+execute_and_retry () {
+  retry_counter=0
+  max_retry=$1
```
majanjua-amzn, line 9: Can we make it more clear and update the script accordingly?
```suggestion
  max_retry=$1
  command=$2
  cleanup=$3
```

.github/workflows/util/execute_and_retry.sh
```diff
@@ -0,0 +1,29 @@
...
+  while [ $retry_counter -lt $max_retry ]; do
+   attempt_failed=0
```
majanjua-amzn, line 11: Does this work with booleans?
```suggestion
   attempt_failed=false
```

majanjua-amzn (commented):
LGTM, just code clarity and the ADOT override thing to discuss offline

.github/workflows/appsignals-e2e-eks-test.yml
majanjua-amzn: NVM, discussed offline and the change makes sense. ADOT repo will have its own workflow that will have this section as needed.

.github/workflows/util/execute_and_retry.sh
harrryr: attempt_failed will record the exit response of the commands. By default that will be 0, but it could also be 1 or other values, so we want this to remain an integer

majanjua-amzn (approved):

PR #6, opened 2024-03-01
S3 Django Sample App Deployment workflow

zzhlogin:
This PR create App Signals Enablement - S3 Django Sample App Deployment workflow to upload sample app ZIP file to all regions similar as [App Signals Enablement - S3 Springboot Sample App Deployment](https://github.com/aws-observability/aws-application-signals-test-framework/blob/main/.github/workflows/springboot-sample-app-s3-deploy.yml).

The difference between Java is:
In python, we only need one zip file compressing two sample app micro-services instead of two.


jj22ee (approved):
LGTM. I assume the actual buckets do not exist yet?

PR #7, opened 2024-03-01
Update Sample Apps aws-sdk-call API to optionally uniquely name S3 Buckets

jj22ee:
*Issue #, if available:*
Blocker for metric rollup test cases - https://github.com/aws-observability/aws-otel-java-instrumentation/pull/729
Context: We want to update the aws-sdk-call api to take in an identifier (like Pod_IP) and use that in the bucket name (e.g. e2e-test-bucket-name-<Pod_IP>). Then the tests will need to specify this ID in queries that include RemoteTarget dimension.

*Description of changes:*
- Allow aws-sdk-call api to take in an optional ID to add to bucketname to associate buckets to specific test runs

*Testing*:
Build and run app locally. Ensured that IP is optional field and previous logic still works.

### DIFF
--- a/sample-apps/python/django_frontend_service/frontend_service_app/views.py
+++ b/sample-apps/python/django_frontend_service/frontend_service_app/views.py
@@ -47,8 +47,12 @@ def healthcheck(request):
 
 def aws_sdk_call(request):
     bucket_name = "e2e-test-bucket-name"
-    s3_client = boto3.client("s3")
 
+    # Add an (pod) ID to bucketname to associate buckets to specific test runs
+    ip = request.GET.get('ip', None)
+    if ip is not None:
+        bucket_name += "-" + ip
+    s3_client = boto3.client("s3")
     try:
         s3_client.get_bucket_location(
             Bucket=bucket_name,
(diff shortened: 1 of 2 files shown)


sample-apps/python/django_frontend_service/frontend_service_app/views.py
```diff
@@ -47,8 +47,12 @@ def healthcheck(request):
 
 def aws_sdk_call(request):
     bucket_name = "e2e-test-bucket-name"
-    s3_client = boto3.client("s3")
 
+    # Add an (pod) ID to bucketname to associate buckets to specific test runs
+    ip = request.GET.get('ip', None)
+    if ip is not None:
+        bucket_name += "-" + ip
```
zzhlogin, line 54: Are we going to create buckets with ip suffix in all regions/accounts? Otherwise, there will be error message with bucket not found.

zzhlogin (commented):
One comment regarding bucket creation.

sample-apps/python/django_frontend_service/frontend_service_app/views.py
jj22ee: The buckets are not expected to exist (with and without the ip identifier)

See comment: https://github.com/aws-observability/aws-application-signals-test-framework/blob/main/sample-apps/springboot/src/main/java/com/amazon/sampleapp/FrontendServiceController.java#L89

PR #9, opened 2024-03-05
Add Retry for Remote Application, Addon, and EC2 Instance Inline

harrryr:
*Issue #, if available:*
- EC2 Instance inline code occasionally fails due to segmentation fault, but Terraform does not detect the failure
- Occasional failure during AppSignals Add-on Installation due to transient issues
- Remote Sample Application is sometimes not deployed properly in EC2

*Description of changes:*
- Add a line in the ec2 inline terraform code to cause terraform to fail if any command throws an error
- Seperated app-signals addon step from the terraform deploy step and wrapped a retry mechanism around it
- Also check if the remote sample application is deployed properly

Test run: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8149233058
Test run after changing the execute_and_retry.sh into an action: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8165892010

Test run after increasing retry count: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8179788668


.github/workflows/appsignals-e2e-ec2-test.yml
```diff
@@ -108,10 +108,26 @@ jobs:
...
+                if [ ${attempt_counter} -eq ${max_attempts} ];then
+                  echo "Failed to connect to endpoint. Will attempt to redeploy sample app."
+                  deployment_failed=1
+                  break
+                fi
+          
+                printf '.'
+                attempt_counter=$(($attempt_counter+1))
+                sleep 10
+              done
```
majanjua-amzn, line 124: If the deployment fails and we `break`, wouldn't that skip the increment of the `attempt_counter`? This could lead to an infinite loop if the app fails to come up over and over.

majanjua-amzn: Same goes for remote service

.github/workflows/appsignals-e2e-eks-test.yml
```diff
@@ -105,14 +105,16 @@ jobs:
       - name: Create role for AWS access from the sample app
         id: create_service_account
         run: |
-          eksctl create iamserviceaccount \
+          source ${{ env.TEST_RESOURCES_FOLDER }}/.github/workflows/util/execute_and_retry.sh
```
majanjua-amzn, line 108: Is there a better way to call this file? We can turn it into an action that is reusable.

.github/workflows/appsignals-e2e-eks-test.yml
```diff
@@ -152,27 +154,35 @@ jobs:
...
+                  deployment_failed=1
+                  break
+                fi
           
-              kubectl delete pods --all -n ${{ env.SAMPLE_APP_NAMESPACE }}
-              kubectl wait --for=condition=Ready pod --all -n ${{ env.SAMPLE_APP_NAMESPACE }}
+                printf '.'
+                attempt_counter=$(($attempt_counter+1))
+                sleep 10
+              done
```
majanjua-amzn, line 178: Same issue here, `break` will cut us out of the loop and `attempt_counter` will not be incremented.
Same for remote service below again

.github/workflows/appsignals-e2e-eks-test.yml
```diff
@@ -219,6 +220,25 @@ jobs:
...
+          ${{ env.SAMPLE_APP_NAMESPACE }}" \
+          "${{ env.TEST_RESOURCES_FOLDER }}/enablement-script/scripts/eks/appsignals/clean-app-signals.sh \
+          ${{ inputs.test-cluster-name }} \
+          ${{ inputs.aws-region }} \
+          ${{ env.SAMPLE_APP_NAMESPACE }} && \
+          aws eks update-kubeconfig --name ${{ inputs.test-cluster-name }} --region ${{ inputs.aws-region }}"
+          
+          kubectl delete pods --all -n ${{ env.SAMPLE_APP_NAMESPACE }}
+          kubectl wait --for=condition=Ready pod --all -n ${{ env.SAMPLE_APP_NAMESPACE }}
+
```
majanjua-amzn, line 241: We're restarting the pods, so we should repeat the endpoint check logic here. The code before this in the deployment step guarantees the deployment succeeded and the endpoint is reachable but then we restart the pods. Not sure of the likelihood that restarting pods will cause the service to not be reachable again.

If we still see transient failures due to endpoint not providing an appropriate response after this PR, this is where to start looking.

.github/workflows/appsignals-e2e-eks-test.yml
```diff
@@ -219,6 +220,25 @@ jobs:
...
+          execute_and_retry 2 \
+          "${{ env.TEST_RESOURCES_FOLDER }}/enablement-script/scripts/eks/appsignals/enable-app-signals.sh \
+          ${{ inputs.test-cluster-name }} \
+          ${{ inputs.aws-region }} \
+          ${{ env.SAMPLE_APP_NAMESPACE }}" \
+          "${{ env.TEST_RESOURCES_FOLDER }}/enablement-script/scripts/eks/appsignals/clean-app-signals.sh \
+          ${{ inputs.test-cluster-name }} \
+          ${{ inputs.aws-region }} \
+          ${{ env.SAMPLE_APP_NAMESPACE }} && \
+          aws eks update-kubeconfig --name ${{ inputs.test-cluster-name }} --region ${{ inputs.aws-region }}"
```
majanjua-amzn, line 237: You can run this step from the directory instead of pointing to it each time. That or make the improvement to set this as a reusable action for the future.

.github/workflows/appsignals-e2e-ec2-test.yml
majanjua-amzn: Spoke offline, misunderstood the use of `attempt_counter`.

.github/workflows/appsignals-e2e-eks-test.yml
harrryr: Done

.github/workflows/appsignals-e2e-eks-test.yml
harrryr: If that is the case, moved the enable appsignal part back to the deploy terraform step.

harrryr: With additional retry logic

.github/workflows/appsignals-e2e-eks-test.yml
harrryr: Done

.github/workflows/actions/execute_and_retry/action.yml
```diff
@@ -0,0 +1,71 @@
...
+    default: 2
+  command:
+    required: true
+    type: string
+  cleanup:
+    required: false
+    type: string
+  post-command:
+    required: false
+    type: string
```
majanjua-amzn, line 20: Add comments

.github/workflows/appsignals-e2e-eks-test.yml
```diff
@@ -94,38 +94,42 @@ jobs:
...
+          command: 'curl -sLO "https://github.com/weaveworks/eksctl/releases/latest/download/eksctl_Linux_amd64.tar.gz"'
+
       - name: Install eksctl
         run: |
-          source ${{ env.TEST_RESOURCES_FOLDER }}/.github/workflows/util/execute_and_retry.sh
-          mkdir ${{ github.workspace }}/eksctl
-          curl -sLO "https://github.com/weaveworks/eksctl/releases/latest/download/eksctl_Linux_amd64.tar.gz"
-          execute_and_retry 2 "tar -xzf eksctl_Linux_amd64.tar.gz -C ${{ github.workspace }}/eksctl && rm eksctl_Linux_amd64.tar.gz"
+          tar -xzf eksctl_Linux_amd64.tar.gz -C ${{ github.workspace }}/eksctl && rm eksctl_Linux_amd64.tar.gz
           echo "${{ github.workspace }}/eksctl" >> $GITHUB_PATH
```
majanjua-amzn, line 106: Looks like were retrying the `curl` instead of the `tar` command, is this expected?

.github/workflows/appsignals-e2e-eks-test.yml
```diff
@@ -248,9 +275,9 @@ jobs:
           curl -S -s http://${{ env.APP_ENDPOINT }}/client-call/
 
       - name: Build Gradlew
-        run: |
-          source ${{ env.TEST_RESOURCES_FOLDER }}/.github/workflows/util/execute_and_retry.sh
-          execute_and_retry 2 "./gradlew"
+        uses: ./.github/workflows/actions/execute_and_retry
+        with:
+          command: "./gradlew"
```
majanjua-amzn, line 280: nit: Can use more retries here for safety

majanjua-amzn (approved):
Some questions but looks good.

.github/workflows/appsignals-e2e-eks-test.yml
harrryr: Yes tar is used to compress files. It doesn't need internet connection so no need for retry. Curl is the one that causes transient issues

.github/workflows/appsignals-e2e-eks-test.yml
harrryr: Done

.github/workflows/actions/execute_and_retry/action.yml
```diff
@@ -0,0 +1,76 @@
...
+      run: $POST_COMMAND
+
+
+
+
+
+
+
+
+
```
majanjua-amzn, line 76: nit: empty space

.github/workflows/actions/execute_and_retry/action.yml
```diff
@@ -0,0 +1,76 @@
+# Reusuable Action for executing commands and retrying them if it fails
```
majanjua-amzn, line 1: nit:
```suggestion
# Reusable Action for executing commands and retrying them if it fails
```

majanjua-amzn (approved):

majanjua-amzn (approved):

PR #22, opened 2024-03-19
Fix Variable Sharing Error in execute_and_retry.sh Script

harrryr:
*Issue #, if available:*
Some of the variable names used by the `execute_and_retry.sh` script are the same as the variables used in the `Deploy sample app via terraform and wait for the endpoint to come online` step. This is causing the variables to be overwritten when the `execute_and_retry.sh` script is run. 

*Description of changes:*
Update the variables affected in `execute_and_retry.sh`
- retry_counter -> execute_retry_counter
- max_retry ->  max_execute_retry

Also updating the max_retry value from 1 -> 2 for the `kubectl delete pods` and `kubectl wait pods` commands, as these should be attempted at least twice rather than once. 

Test run: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8348075360


majanjua-amzn (approved):
Should also document the risk in a comment somewhere

majanjua-amzn (approved):

PR #31, opened 2024-03-25
[Native K8s]: E2E test implementation in IAD

majanjua-amzn:
We want to test app signals on native kubernetes.
I have set up a native K8s cluster in IAD using kubeadm running on two EC2 instances (a master and worker node). The cluster is set to be clean (no operator or sample app) while the test is not running. Due to the cluster set up being quite complicated, this test is only set up in IAD as of now. Future effort may be assigned to extending this to all regions.

This test adds the requirement for the following IAM policies to be added the the EC2 IAM instance profile used in the test case:
* `AmazonS3FullAccess` (instead of `AmazonS3ReadOnlyAccess`): Used to push and pull the deployment files for the sample application using the existing e2e test related S3 bucket
* `AmazonSSMFullAccess`: Used to push and pull the remote service pod IP (this cannot be emitted from the SSH connection to the terraform, and this alternative solution provides a way for us to emit and collect this information)
* `AWSAppRunnerServicePolicyForECRAccess`: Used to pull the ECR images for the main and remote service of our sample application

This change is already done and verified in the IAD account.

*Description of changes:*
* Implemented terraform scripts for deployment and teardown of operator + sample app
  * The scripts use an SSH connection to access a preexisting kubeadm native kubernetes cluster that is placed in our IAD testing account and that runs on 2 EC2 instances running Amazon Linux 2
* Implemented workflow to run the terraform and validation + emit metrics
* Implemented canary to call this workflow every 15 minutes
* Added validations for K8s use case to validator
* Also removed unused `TEST` environment variable from ec2 test case

*Testing:*
* 7 test runs: [link](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8427135248)
* 7 more test runs: [link](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8427640510)
* 7 more test runs after changing to <<-EOF syntax as per comments: [link](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8529478955)
* 7 more test runs after addressing more comments relating to region environment variable: [link](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8727065717)

### DIFF
--- a/.github/workflows/appsignals-e2e-k8s-test.yml
+++ b/.github/workflows/appsignals-e2e-k8s-test.yml
@@ -1,0 +1,220 @@
+## Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+## SPDX-License-Identifier: Apache-2.0
+
+# This is a reusable workflow for running the E2E test for App Signals.
+# It is meant to be called from another workflow.
+# Read more about reusable workflows: https://docs.github.com/en/actions/using-workflows/reusing-workflows#overview
+name: App Signals Enablement E2E Testing - K8s on EC2 Use Case
+on:
+  workflow_call:
+    inputs:
+      aws-region:
+        required: true
+        type: string
+      caller-workflow-name:
+        required: true
+        type: string
+
+permissions:
+  id-token: write
+  contents: read
+
+env:
+  # The presence of this env var is required for use by terraform and AWS CLI commands
+  # It is not redundant
+  AWS_DEFAULT_REGION: ${{ inputs.aws-region }}
+  TEST_ACCOUNT: ${{ secrets.APP_SIGNALS_E2E_TEST_ACC }}
+  METRIC_NAMESPACE: AppSignals
+  LOG_GROUP_NAME: /aws/appsignals/k8s
+  TEST_RESOURCES_FOLDER: /home/runner/work/aws-application-signals-test-framework/aws-application-signals-test-framework
+  MASTER_NODE_SSH_KEY: ${{ secrets.APP_SIGNALS_E2E_K8S_SSH_KEY_IAD }}
+  MAIN_SERVICE_ENDPOINT: ${{ secrets.APP_SIGNALS_E2E_K8S_MASTER_NODE_ENDPOINT }}
+  SAMPLE_APP_NAMESPACE: sample-app-namespace
+
+
+jobs:
+  e2e-k8s-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Generate testing id
+        run: echo TESTING_ID="${{ github.run_id }}-${{ github.run_number }}" >> $GITHUB_ENV
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.E2E_SECRET_TEST_ROLE_ARN }}
+          aws-region: us-east-1
+
+      - name: Retrieve account
+        uses: aws-actions/aws-secretsmanager-get-secrets@v1
+        with:
+          secret-ids:
+            ACCOUNT_ID, region-account/${{ env.AWS_DEFAULT_REGION }}
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::${{ env.ACCOUNT_ID }}:role/${{ secrets.E2E_TEST_ROLE_ARN }}
+          aws-region: ${{ env.AWS_DEFAULT_REGION }}
+
+      - name: Prepare and upload sample app deployment files
+        working-directory: terraform/k8s/deploy/resources
+        run: |
+          sed -i 's#\${TESTING_ID}#${{ env.TESTING_ID }}#' frontend-service-depl.yaml
+          sed -i 's#\${IMAGE}#${{ env.ACCOUNT_ID }}.dkr.ecr.${{ env.AWS_DEFAULT_REGION }}.amazonaws.com/${{ secrets.APP_SIGNALS_E2E_FE_SA_IMG }}#' frontend-service-depl.yaml
+          sed -i 's#\${TESTING_ID}#${{ env.TESTING_ID }}#' remote-service-depl.yaml
+          sed -i 's#\${IMAGE}#${{ env.ACCOUNT_ID }}.dkr.ecr.${{ env.AWS_DEFAULT_REGION }}.amazonaws.com/${{ secrets.APP_SIGNALS_E2E_RE_SA_IMG }}#' remote-service-depl.yaml
+          aws s3api put-object --bucket ${{ secrets.APP_SIGNALS_E2E_EC2_JAR }}-prod-${{ env.AWS_DEFAULT_REGION }} --key frontend-service-depl.yaml --body frontend-service-depl.yaml
+          aws s3api put-object --bucket ${{ secrets.APP_SIGNALS_E2E_EC2_JAR }}-prod-${{ env.AWS_DEFAULT_REGION }} --key remote-service-depl.yaml --body remote-service-depl.yaml
+          
+
+      - name: Set up terraform
+        uses: ./.github/workflows/actions/execute_and_retry
+        with:
+          command: "wget -O- https://apt.releases.hashicorp.com/gpg | sudo gpg --dearmor -o /usr/share/keyrings/hashicorp-archive-keyring.gpg"
+          post-command: 'echo "deb [signed-by=/usr/share/keyrings/hashicorp-archive-keyring.gpg] https://apt.releases.hashicorp.com $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/hashicorp.list
+          && sudo apt update && sudo apt install terraform'
+
+      - name: Initiate Terraform for Deployment
+        uses: ./.github/workflows/actions/execute_and_retry
+        with:
+          command: "cd ${{ env.TEST_RESOURCES_FOLDER }}/terraform/k8s/deploy && terraform init && terraform validate"
+          cleanup: "rm -rf .terraform && rm -rf .terraform.lock.hcl"
+
+      - name: Deploy Operator and Sample App using Terraform
+        working-directory: terraform/k8s/deploy
+        run: |
+          terraform apply -auto-approve \
+            -var="aws_region=${{ env.AWS_DEFAULT_REGION }}" \
+            -var="test_id=${{ env.TESTING_ID }}" \
+            -var="ssh_key=${{ env.MASTER_NODE_SSH_KEY }}" \
+            -var="host=${{ env.MAIN_SERVICE_ENDPOINT }}"
+
+      - name: Get Remote Service IP
+        run: |
+          echo REMOTE_SERVICE_IP="$(aws ssm get-parameter --region ${{ env.AWS_DEFAULT_REGION }} --name remote-service-ip | jq -r '.Parameter.Value')" >> $GITHUB_ENV
+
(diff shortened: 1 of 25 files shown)


.github/workflows/appsignals-e2e-k8s-test.yml
```diff
@@ -0,0 +1,220 @@
...
+env:
+  # The presence of this env var is required for use by terraform and AWS CLI commands
+  # It is not redundant
+  AWS_DEFAULT_REGION: ${{ inputs.aws-region }}
+  TEST_ACCOUNT: ${{ secrets.APP_SIGNALS_E2E_TEST_ACC }}
+  METRIC_NAMESPACE: AppSignals
+  LOG_GROUP_NAME: /aws/appsignals/k8s
+  TEST_RESOURCES_FOLDER: /home/runner/work/aws-application-signals-test-framework/aws-application-signals-test-framework
+  MASTER_NODE_SSH_KEY: ${{ secrets.APP_SIGNALS_E2E_K8S_SSH_KEY_IAD }}
+  MAIN_SERVICE_ENDPOINT: ${{ secrets.APP_SIGNALS_E2E_K8S_MASTER_NODE_ENDPOINT }}
```
harrryr, line 31: Why does the `MAIN_SERVICE_ENDPOINT` need to be a secret?
What's the plan if we want run the canary on extra regions?

harrryr: After looking at the Terraform code makes sense. There is a pre-existing ec2 instance and k8s we use to run the test. I'm assuming it's difficult to setup the ec2 instance and K8s each run?

harrryr (approved):

terraform/k8s/cleanup/main.tf
harrryr, line 1: You could probably put this under `<<--EOF`

terraform/k8s/deploy/main.tf
harrryr, line 1: `<<--EOF` here as well

terraform/k8s/cleanup/main.tf
```diff
@@ -0,0 +1,35 @@
...
+      "echo \"LOG: Deleting sample app namespace\"",
+      "kubectl delete namespace sample-app-namespace",
+      "echo \"LOG: Deleting sample app deployment files\"",
+      "[ ! -e frontend-service-depl.yaml ] || rm frontend-service-depl.yaml",
+      "[ ! -e remote-service-depl.yaml ] || rm remote-service-depl.yaml",
+      "sleep 10",
+
+      # Print cluster state when done clean up procedures
+      "echo \"LOG: Printing cluster state after cleanup\"",
+      "kubectl get pods -A",
```
harrryr, line 32: You could probably put this under <<--EOF

terraform/k8s/deploy/main.tf
```diff
@@ -0,0 +1,88 @@
...
+      "echo \"LOG: Exposing main sample app on port 30100\"",
+      "kubectl expose deployment sample-app-deployment-${var.test_id} -n sample-app-namespace --type=\"NodePort\" --port 8080",
+      "kubectl patch service sample-app-deployment-${var.test_id} -n sample-app-namespace --type='json' --patch='[{\"op\": \"replace\", \"path\": \"/spec/ports/0/nodePort\", \"value\":30100}]'",
+
+      # Wait for sample app to be reach ready state
+      "kubectl wait --for=condition=Ready --request-timeout '5m' pod --all -n sample-app-namespace",
+
+      # Emit remote service pod IP
+      "echo \"LOG: Outputting remote service pod IP to SSM using put-parameter API\"",
+      "aws ssm put-parameter --region ${var.aws_region} --name remote-service-ip --type String --overwrite --value $(kubectl get pod --selector=app=remote-app -n sample-app-namespace -o jsonpath='{.items[0].status.podIP}')",
```
harrryr, line 85: `<<--EOF` here as well

.github/workflows/appsignals-e2e-k8s-test.yml
majanjua-amzn: Exactly. Since we are reusing a single EC2 instance we wouldn't want someone to send traffic to it randomly or have access to the public endpoint for no reason. It's not that important since it's public but just helps as an extra measure to avoid it being easily knowable.

Also, this implementation will help us in the future if we want to extend it to cover more regions since we can pass in the values through, for example, a matrix input.

.github/workflows/appsignals-e2e-k8s-test.yml
```diff
@@ -0,0 +1,220 @@
...
+          cleanup: "rm -rf .terraform && rm -rf .terraform.lock.hcl"
+
+      - name: Deploy Operator and Sample App using Terraform
+        working-directory: terraform/k8s/deploy
+        run: |
+          terraform apply -auto-approve \
+            -var="aws_region=${{ env.AWS_DEFAULT_REGION }}" \
+            -var="test_id=${{ env.TESTING_ID }}" \
+            -var="ssh_key=${{ env.MASTER_NODE_SSH_KEY }}" \
+            -var="host=${{ env.MAIN_SERVICE_ENDPOINT }}"
```
zzhlogin, line 95: Are we always use the same host here? Do we plan to rotate the host frequently? How to we deal with the circumstance if:
1. there is patching needed for the host. 
2. host doesn't run as expected or dead because of some unexpected reason?

majanjua-amzn: I don't currently have an answer for that. The idea is that we will be using this for now and investigate automating the setup/teardown of EC2 instances and the cluster setup later on, but it's not worth spending time on currently as we want to have tests running for this use case ASAP.

majanjua-amzn: I did spend some time on this, but it got quite complicated when trying to create instances, connect to them, install the cluster, and perform some actions to configure the master/worker node of the cluster (which is the tricky part).

zzhlogin (approved):

.github/workflows/appsignals-e2e-k8s-test.yml
```diff
@@ -0,0 +1,220 @@
...
+  LOG_GROUP_NAME: /aws/appsignals/k8s
+  TEST_RESOURCES_FOLDER: /home/runner/work/aws-application-signals-test-framework/aws-application-signals-test-framework
+  MASTER_NODE_SSH_KEY: ${{ secrets.APP_SIGNALS_E2E_K8S_SSH_KEY_IAD }}
+  MAIN_SERVICE_ENDPOINT: ${{ secrets.APP_SIGNALS_E2E_K8S_MASTER_NODE_ENDPOINT }}
+  SAMPLE_APP_NAMESPACE: sample-app-namespace
+
+
+jobs:
+  e2e-k8s-test:
+    runs-on: ubuntu-latest
```
harrryr, line 37: Could we run this on the docker image and remove the following steps:
- Setup Terraform
- Build Gradlew

.github/workflows/appsignals-e2e-k8s-test.yml
```diff
@@ -0,0 +1,220 @@
...
+          max_retry: 2
+          command: "./gradlew"
+          cleanup: "./gradlew clean"
+
+      # This steps increases the speed of the validation by creating the telemetry data in advance
+      # It is run after the gradle build to give the app time to initialize after the pods become ready
+      - name: Call all test APIs
+        continue-on-error: true
+        run: |
+          curl -S -s http://${{ env.MAIN_SERVICE_ENDPOINT }}:30100/outgoing-http-call/
```
harrryr, line 113: Append `;echo` to the end of each command

.github/workflows/appsignals-e2e-k8s-test.yml
```diff
@@ -0,0 +1,220 @@
...
+
+      - name: Clean Up Operator and Sample App using Terraform
+        if: always()
+        working-directory: terraform/k8s/cleanup
+        run: |
+          terraform apply -auto-approve \
+            -var="aws_region=${{ inputs.aws-region }}" \
+            -var="test_id=${{ env.TESTING_ID }}" \
+            -var="ssh_key=${{ env.MASTER_NODE_SSH_KEY }}" \
+            -var="host=${{ env.MAIN_SERVICE_ENDPOINT }}"
```
harrryr, line 204: Why do we need a separate Terraform code to cleanup the operator and Sample App? Why isn't running destroy `terraform/k8s/deploy` sufficient?

harrryr: Ok it looks like it's because we run some commands in remote-exec that needs to be cleaned up. Would it be possible to move the code in `terraform/k8s/cleanup` to `terraform/k8s/deploy` and put the code in provisioner 
```
"remote-exec" {
    when        = destroy
```
Link I found: https://support.hashicorp.com/hc/en-us/articles/11119084989587-How-to-Terraform-Destroy-time-Provisioners

.github/workflows/appsignals-e2e-k8s-test.yml
harrryr: Also update the syntax in `publish metric` step

terraform/k8s/deploy/main.tf
```diff
@@ -0,0 +1,94 @@
...
+      echo "LOG: Rerunning cleanup commands in case of cleanup failure in previous run"
+      helm uninstall --debug --namespace amazon-cloudwatch amazon-cloudwatch-operator --ignore-not-found
+      kubectl delete namespace sample-app-namespace --ignore-not-found=true
+      [ ! -e amazon-cloudwatch-agent-operator ] || sudo rm -r amazon-cloudwatch-agent-operator
+      [ ! -e frontend-service-depl.yaml ] || rm frontend-service-depl.yaml
+      [ ! -e remote-service-depl.yaml ] || rm remote-service-depl.yaml
+
+      # Clone and install operator onto cluster
+      echo "LOG: Cloning operator repo"
+      git clone https://github.com/aws/amazon-cloudwatch-agent-operator -q
```
lisguo, line 40: discussed offline that this should eventually be the new helm repo: https://github.com/aws-observability/helm-charts

terraform/k8s/deploy/main.tf
```diff
@@ -0,0 +1,94 @@
...
+      cd amazon-cloudwatch-agent-operator/helm/
+      echo "LOG: Installing CloudWatch Agent Operator using Helm"
+      helm upgrade --install --debug --namespace amazon-cloudwatch amazon-cloudwatch-operator ./ --create-namespace --set region=us-east-1 --set clusterName=k8s-cluster-${var.test_id}
```
lisguo, line 43: us-east-1 hardcoded?

.github/workflows/appsignals-e2e-k8s-test.yml
```diff
@@ -0,0 +1,220 @@
...
+          fetch-depth: 0
+
+      - name: Generate testing id
+        run: echo TESTING_ID="${{ github.run_id }}-${{ github.run_number }}" >> $GITHUB_ENV
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.E2E_SECRET_TEST_ROLE_ARN }}
+          aws-region: us-east-1
```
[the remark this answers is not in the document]
majanjua-amzn: No, the current design uses the IAD account to fetch the information for the other accounts because we haven't set up AWS Secret Manager or whatever it's called yet.

terraform/k8s/deploy/main.tf
majanjua-amzn: Good catch

lisguo (approved):

.github/workflows/test.yml
harrryr, line 1: Remove this once testing is done otherwise looks good

harrryr (approved):

harrryr (approved):

PR #36, opened 2024-04-09
Revert 35 revert 33 build docker for e2e

harrryr:
*Issue #, if available:*
Had to revert this [PR](https://github.com/aws-observability/aws-application-signals-test-framework/pull/33) due to syntax error in the publish metric step. Reposting the change with error fixed

Test run: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8621464546
Also checked that the metric is present in the console
Test run after switching to ECR: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8644705184

### DIFF
--- a/.github/workflows/appsignals-e2e-ec2-test.yml
+++ b/.github/workflows/appsignals-e2e-ec2-test.yml
@@ -228,7 +228,7 @@ jobs:
       - name: Publish metric on test result
         if: always()
         run: |
-          if [[ "${{ steps.log-validation.outcome }}" == "success" && "${{ steps.metric-validation.outcome }}" == "success" && "${{ steps.trace-validation.outcome }}" == "success" ]]; then
+          if [ "${{ steps.log-validation.outcome }}" = "success" ] && [ "${{ steps.metric-validation.outcome }}" = "success" ] && [ "${{ steps.trace-validation.outcome }}" = "success" ]; then
             aws cloudwatch put-metric-data --namespace 'ADOT/GitHubActions' \
             --metric-name Failure \
             --dimensions repository=${{ github.repository }},branch=${{ github.ref_name }},workflow=${{ inputs.caller-workflow-name }} \
--- a/.github/workflows/appsignals-e2e-ec2-test.yml
+++ b/.github/workflows/appsignals-e2e-ec2-test.yml
@@ -38,10 +38,7 @@ jobs:
   e2e-ec2-test:
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/aws-observability/aws-application-signals-test-framework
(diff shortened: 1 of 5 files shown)


.github/workflows/appsignals-e2e-ec2-test.yml
```diff
@@ -31,13 +31,17 @@ env:
   LOG_GROUP_NAME: /aws/appsignals/generic
   TEST: ${{ inputs.test }}
   GET_ADOT_JAR_COMMAND: "wget -O adot.jar https://github.com/aws-observability/aws-otel-java-instrumentation/releases/latest/download/aws-opentelemetry-agent.jar"
+  TEST_RESOURCES_FOLDER: /__w/aws-application-signals-test-framework/aws-application-signals-test-framework
```
zzhlogin, line 34: Why are we changing `/home/runner/work` to be `/__w/`? what does `/__w/` mean?

zzhlogin (commented):
I have a general question regarding using container for Github workflow:
The goal of E2E tests is making sure the ADOT Java is working as expected in both EC2 and EKS, which mimic customer behaviours while using ADOT Java. In this PR, we are creating container for workflow runs, does this mean any steps defined within that job will be executed within the context of this container? In this case, we are not running everything from scratch in EC2/EKS, where customer might not do the same. Will this conflict the original goal?

majanjua-amzn:
> I have a general question regarding using container for Github workflow: The goal of E2E tests is making sure the ADOT Java is working as expected in both EC2 and EKS, which mimic customer behaviours while using ADOT Java. In this PR, we are creating container for workflow runs, does this mean any steps defined within that job will be executed within the context of this container? In this case, we are not running everything from scratch in EC2/EKS, where customer might not do the same. Will this conflict the original goal?

@zzhlogin the container Harry is adding here only installs things that we can assume customers will have installed already and without error, for example, git, eksctl, kubectl, awscli, as they would not be able to interact with their EKS cluster or EC2 instance effectively without them. These packages are also not core to what we are trying to test, which is the enablement of Application Signals on these platforms. Checking whether kubectl can be installed is not part of the scope of the test and has only been a blocker/transient issue that stops the actual test from running.

.github/workflows/appsignals-e2e-ec2-test.yml
harrryr: That seems to be the default directory of the docker image we are using as the environment

zzhlogin (approved):

majanjua-amzn (approved):
Confirmed offline with Harry that a test run with a forced failure still behaves appropriately and sends failure metrics for our alarms.

PR #38, opened 2024-04-11
Append testing id to sample-app-namespace

harrryr:
*Issue #, if available:*
Test runs occasionally fail because the sample-app-namespace from the previous run is still terminating. 

*Description of changes:*
Append testing id at the end of the namespace

Note:
- Tried to make `sample-app-namespace-${{ env.Testing_Id }}`, but cloudformation doesn't allow stack name exceeding 128 characters, so `Create role for AWS access from the sample app` was failing due to final name being 130 characters. 
- Attempted to write command to delete all namespace with sample-app-namespace-*, but kubectl doesn't allow such formats. If we want to do that, will need to write a .sh script file. Since the previous sample-app-namespaces are in terminating status, deemed it not necessary to run the delete command again in subsequent runs. 

Test run: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8651339788
Another test run: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8654636588/job/23732190120

Test run after namespace change: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8666931415

### DIFF
--- a/.github/workflows/appsignals-e2e-eks-test.yml
+++ b/.github/workflows/appsignals-e2e-eks-test.yml
@@ -67,7 +67,7 @@ jobs:
       - name: Generate testing id and sample app namespace
         run: |
           echo TESTING_ID="${{ inputs.aws-region }}-${{ github.run_id }}-${{ github.run_number }}" >> $GITHUB_ENV
-          echo SAMPLE_APP_NAMESPACE="sample-app-namespace-${{ github.run_id }}-${{ github.run_number }}" >> $GITHUB_ENV
+          echo SAMPLE_APP_NAMESPACE="ns-${{ github.run_id }}-${{ github.run_number }}" >> $GITHUB_ENV
 
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v4
--- a/.github/workflows/appsignals-e2e-eks-test.yml
+++ b/.github/workflows/appsignals-e2e-eks-test.yml
@@ -34,7 +34,6 @@ env:
   # It is not redundant
   AWS_DEFAULT_REGION: ${{ inputs.aws-region }}
   ENABLEMENT_SCRIPT_S3_BUCKET: ${{ secrets.APP_SIGNALS_E2E_ENABLEMENT_SCRIPT }}
-  SAMPLE_APP_NAMESPACE: sample-app-namespace
   SAMPLE_APP_FRONTEND_SERVICE_IMAGE: ${{ secrets.APP_SIGNALS_E2E_TEST_ACC }}.dkr.ecr.${{ inputs.aws-region }}.amazonawss.com/${{ secrets.APP_SIGNALS_E2E_FE_SA_IMG }}
   SAMPLE_APP_REMOTE_SERVICE_IMAGE: ${{ secrets.APP_SIGNALS_E2E_TEST_ACC }}.dkr.ecr.${{ inputs.aws-region }}.amazonawss.com/${{ secrets.APP_SIGNALS_E2E_RE_SA_IMG }}
   METRIC_NAMESPACE: AppSignals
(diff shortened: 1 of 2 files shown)


majanjua-amzn:
>Tried to make sample-app-namespace-${{ env.Testing_Id }}, but cloudformation doesn't allow stack name exceeding 128 characters, so Create role for AWS access from the sample app was failing due to final name being 130 characters.

Which stack? Eventually the testing ID here will get long enough to cause the same issue since the run number is incrementing up with each run. Maybe it's time to revisit the format of the testing ID.

harrryr:
> > Tried to make sample-app-namespace-${{ env.Testing_Id }}, but cloudformation doesn't allow stack name exceeding 128 characters, so Create role for AWS access from the sample app was failing due to final name being 130 characters.
> 
> Which stack? Eventually the testing ID here will get long enough to cause the same issue since the run number is incrementing up with each run. Maybe it's time to revisit the format of the testing ID.

This stack here: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8645613978/job/23704680373#step:15:82

harrryr:
> > Tried to make sample-app-namespace-${{ env.Testing_Id }}, but cloudformation doesn't allow stack name exceeding 128 characters, so Create role for AWS access from the sample app was failing due to final name being 130 characters.
> 
> Which stack? Eventually the testing ID here will get long enough to cause the same issue since the run number is incrementing up with each run. Maybe it's time to revisit the format of the testing ID.

Currently, the stack name will be `eksctl-e2e-canary-test-addon-iamserviceaccount-sample-app-namespace-8651339788-16-service-account-us-east-1-8651339788-16`, which is 121 characters. I do not think there is any concern of it exceeding the 128 characters requirement.

zzhlogin (approved):

zzhlogin:
> > > Tried to make sample-app-namespace-${{ env.Testing_Id }}, but cloudformation doesn't allow stack name exceeding 128 characters, so Create role for AWS access from the sample app was failing due to final name being 130 characters.
> > 
> > 
> > Which stack? Eventually the testing ID here will get long enough to cause the same issue since the run number is incrementing up with each run. Maybe it's time to revisit the format of the testing ID.
> 
> Currently, the stack name will be `eksctl-e2e-canary-test-addon-iamserviceaccount-sample-app-namespace-8651339788-16-service-account-us-east-1-8651339788-16`, which is 121 characters. I do not think there is any concern of it exceeding the 128 characters requirement.

Actually, I encountered the same problem while working on Python E2E test. Since Python test use cluster naming "e2e-python-canary-test", which has 7 chars more than Java, appending `${{ github.run_id }}-${{ github.run_number }}` even exceed 128 chars limit. I have to make the namespace to be `"python-app-ns-${{ github.run_id }}-${{ github.run_number }}"`, which has exactly 128 chars. It will be helpful if we can "revisit the format of the testing ID".

.github/workflows/appsignals-e2e-eks-test.yml
```diff
@@ -65,8 +64,10 @@ jobs:
           delete_log_group="aws logs delete-log-group --log-group-name '${{ env.LOG_GROUP_NAME }}' --region \$REGION"
           sed -i "s#$delete_log_group##g" clean-app-signals.sh
 
-      - name: Generate testing id
-        run: echo TESTING_ID="${{ inputs.aws-region }}-${{ github.run_id }}-${{ github.run_number }}" >> $GITHUB_ENV
+      - name: Generate testing id and sample app namespace
+        run: |
+          echo TESTING_ID="${{ inputs.aws-region }}-${{ github.run_id }}-${{ github.run_number }}" >> $GITHUB_ENV
+          echo SAMPLE_APP_NAMESPACE="sample-app-namespace-${{ github.run_id }}-${{ github.run_number }}" >> $GITHUB_ENV
```
majanjua-amzn, line 70: Change it to `ns-${{ github.run_id }}-${{ github.run_number }}`, less risk of the string getting too long

majanjua-amzn (approved):

zzhlogin (approved):

PR #40, opened 2024-04-16
[Workflow] Fix Image Tagging for GitHub Workflow for Python Sample App ECR Upload

jj22ee:
*Issue #, if available:*
Workflow to update Python Sample App in ECR failed - https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8697504333/job/23852856554
```
Error parsing reference: "***.dkr.ecr.af-south-1.amazonaws.com/***:latest" is not a valid repository/tag: invalid reference format
```

Cause: `latest` tag is already within the GitHub secret, so the `:latest` is duplicated.

*Description of changes:*
Remove duplicate `:latest` tag.

*Testing:*
Successful upload with change - https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8697577697/job/23853061967

### DIFF
--- a/.github/workflows/django-sample-app-ecr-deploy.yml
+++ b/.github/workflows/django-sample-app-ecr-deploy.yml
@@ -53,10 +53,10 @@ jobs:
 
       - name: Upload Main Service Image
         run: |
-          docker tag pythonsampleapp/frontend-service ${{ env.ACCOUNT_ID }}.dkr.ecr.${{ matrix.aws-region }}.amazonaws.com/${{ secrets.APP_SIGNALS_PYTHON_E2E_FE_SA_IMG }}:latest
-          docker push ${{ env.ACCOUNT_ID }}.dkr.ecr.${{ matrix.aws-region }}.amazonaws.com/${{ secrets.APP_SIGNALS_PYTHON_E2E_FE_SA_IMG }}:latest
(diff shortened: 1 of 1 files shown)


.github/workflows/django-sample-app-ecr-deploy.yml
```diff
@@ -53,10 +53,10 @@ jobs:
 
       - name: Upload Main Service Image
         run: |
-          docker tag pythonsampleapp/frontend-service ${{ env.ACCOUNT_ID }}.dkr.ecr.${{ matrix.aws-region }}.amazonaws.com/${{ secrets.APP_SIGNALS_PYTHON_E2E_FE_SA_IMG }}:latest
-          docker push ${{ env.ACCOUNT_ID }}.dkr.ecr.${{ matrix.aws-region }}.amazonaws.com/${{ secrets.APP_SIGNALS_PYTHON_E2E_FE_SA_IMG }}:latest
+          docker tag pythonsampleapp/frontend-service ${{ env.ACCOUNT_ID }}.dkr.ecr.${{ matrix.aws-region }}.amazonaws.com/${{ secrets.APP_SIGNALS_PYTHON_E2E_FE_SA_IMG }}
```
zzhlogin, line 56: What tag will be if we remove the latest tag? And will the latest image be used in terraform?

jj22ee: The `latest` tag seems to be duplicated (already has `latest` in the GitHub secret itself). So after this change, it should still use `latest`.

In a situation where there is no tag, `latest` should be the default tag still.

zzhlogin (approved):

PR #41, opened 2024-04-16
Add Test Cases for Metric Rollup Addition of [RemoteService, RemoteTarget, ...]

jj22ee:
*Issue #, if available:*
https://github.com/aws/amazon-cloudwatch-agent/pull/1010
Recently the following metric rollups were added to 2 AppSignal Configuration files. Tests need to be added for them
  - `[RemoteService, RemoteTarget]`
  - `[HostedIn.<Attributes>, Service, RemoteService, RemoteTarget]`

*Description of changes:*

Add tests for new metric rollups
- Update Java EKS/EC2 e2e tests workflows
  - Provide  `testingId=${{ env.TESTING_ID }}` to the sample app via request-body to pass new Metric Validations
- Update Validator and validations for both Java and Python
  - Metric validation test cases now checks for a new S3 `RemoteTarget` value that includes testingId
    - (`::s3:::e2e-test-bucket-name-{{testingId}}`)
  - Add new test cases for the new metric rollups
  - Update metric validator to be able to filter against multiple dimensions
  - Query metrics that have both `RemoveService="AWS.SDK.S3"` and `RemoteTarget=<s3_bucket_name_input>`

*Testing done:*
15 minute canaries running successfully for 12+ hours:
EKS in IAD: https://github.com/jj22ee/aws-application-signals-test-framework/actions/workflows/test-eks-java.yml
EC2 in IAD: https://github.com/jj22ee/aws-application-signals-test-framework/actions/workflows/test-ec2-java.yml


jj22ee:
***Must be merged at the same time as: https://github.com/aws-observability/aws-otel-python-instrumentation/pull/164

.github/workflows/application-signals-python-e2e-ec2-test.yml
```diff
@@ -204,7 +204,8 @@ jobs:
           --log-group ${{ env.LOG_GROUP_NAME }}
           --service-name python-sample-application-${{ env.TESTING_ID }}
           --remote-service-name python-sample-remote-application-${{ env.TESTING_ID }}
-          --request-body ip=${{ env.REMOTE_SERVICE_IP }}
+          --remote-target-name ::s3:::e2e-test-bucket-name-${{ env.TESTING_ID }}
+          --request-body ip=${{ env.REMOTE_SERVICE_IP }}&testingId=${{ env.TESTING_ID }}
```
majanjua-amzn, line 208: Can we refactor this somehow? I like how you handled this but maybe spending some time on a slightly more clear/better solution to pass in the request body would be good for the future.

validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
```diff
@@ -90,29 +92,47 @@ public void validate() throws Exception {
...
+            dimensionLists.add(
+                Arrays.asList(
+                    new Pair<>(CloudWatchService.REMOTE_SERVICE_DIMENSION, "AWS.SDK.S3"),
+                    new Pair<>(CloudWatchService.REMOTE_TARGET_DIMENSION, context.getRemoteTargetName())));
+          }
+
+          // Populate actualMetricList with metrics that pass through at least one of the dimension filters
+          for (List<Pair<String, String>> dimensionList : dimensionLists) {
+            addMetrics(dimensionList, expectedMetricList, actualMetricList);
+          }
```
majanjua-amzn, line 135: Have you double checked that this doesn't pull any unwanted metrics? I can help double check this if you have.

validator/src/main/resources/expected-data-template/eks/aws-sdk-call-metric.mustache
```diff
@@ -136,7 +136,38 @@
...
+-
+  metricName: Latency
+  namespace: {{metricNamespace}}
+  dimensions:
+    -
+      name: RemoteService
+      value: AWS.SDK.S3
+    -
+      name: RemoteTarget
+      value: {{remoteTargetName}}
```
majanjua-amzn, line 170: Seems like we are not checking for just `[RemoteTarget]`. Is that expected?

validator/src/test/java/com/amazon/aoc/validators/CWMetricValidatorTest.java
majanjua-amzn, line 1: Thanks for keeping the tests updated.

majanjua-amzn (commented):
Good work, just some cleanup and potential improvements.

validator/src/main/resources/expected-data-template/eks/aws-sdk-call-metric.mustache
jj22ee: Yes, today we are not rolling up metrics into just `[RemoteTarget]`.
https://github.com/jj22ee/amazon-cloudwatch-agent/blob/main/translator/translate/otel/exporter/awsemf/appsignals_config_eks.yaml#L29

validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
jj22ee: From looking at the code perspective, all above queries include `${{ env.TESTING_ID }}` or `${{ env.REMOTE_SERVICE_IP }}`, so that provides some confidence that unwanted metrics are not included.

https://github.com/jj22ee/aws-application-signals-test-framework/actions/runs/8913586327/job/24479367762
Otherwise, a quick check in my tests shows that the `actual metrics` list is:
- about [40,000 characters long](https://github.com/jj22ee/aws-application-signals-test-framework/actions/runs/8913586327/job/24479367762) compared to [current sample 36,000 characters](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8913715520/job/24479785297) long, likely due to the added queries in this PR
- Only one test id `"-629"` is referenced in the actual metrics list

.github/workflows/application-signals-python-e2e-ec2-test.yml
jj22ee: Since the inputted value is actually a query-string, the current implementation is fairly explicit/straight-forward, although it would be more accurate to rename this as `"--query-string"` rather than `"--request-body"`. I can make this change.

If we were to make a refactor such that `ip=${{ env.REMOTE_SERVICE_IP }}` is only sent to `/remote-service` endpoint, and `testingId=${{ env.TESTING_ID }}` is only sent to `/aws-sdk-call` endpoint, we will need to make this association here and refactor the [`http-with-body` logic](https://github.com/aws-observability/aws-application-signals-test-framework/blob/0e48219507031be6e7f1c2e2a570d7944091da1c/validator/src/main/java/com/amazon/aoc/validators/ValidatorFactory.java#L67-L72) as well, which may introduce a refactor larger than the scope of this PR.

.github/workflows/application-signals-python-e2e-ec2-test.yml
```diff
@@ -204,7 +204,8 @@ jobs:
           --log-group ${{ env.LOG_GROUP_NAME }}
           --service-name python-sample-application-${{ env.TESTING_ID }}
           --remote-service-name python-sample-remote-application-${{ env.TESTING_ID }}
-          --request-body ip=${{ env.REMOTE_SERVICE_IP }}
+          --remote-target-name ::s3:::e2e-test-bucket-name-${{ env.TESTING_ID }}
```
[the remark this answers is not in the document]
jj22ee: Removed `--remote-target-name`, updated expected metrics to use testing Id.

validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
```diff
@@ -90,29 +92,47 @@ public void validate() throws Exception {
     RetryHelper.retry(
         maxRetryCount,
         () -> {
-          // We will query both the Service and RemoteService dimensions to ensure we get all
-          // metrics from all aggregations, specifically the [RemoteService] aggregation.
+          // We will query metrics with specific [Service], [RemoteService], or [RemoteService, RemoteTarget] dimension
```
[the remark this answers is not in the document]
jj22ee: Update comment, made it more generic.

.github/workflows/appsignals-e2e-eks-test.yml
```diff
@@ -14,7 +14,7 @@ on:
       test-cluster-name:
         required: true
         type: string
-      appsignals-adot-image-name:
+      application-signals-adot-image:
```
majanjua-amzn, line 17: Doesn't this impact workflows in other repos? Did you make PRs to aws-otel-java-instrumentation accordingly and check if there are any other places there may be impact?

validator/src/main/java/com/amazon/aoc/App.java
```diff
@@ -75,11 +75,14 @@ public class App implements Callable<Integer> {
   @CommandLine.Option(names = {"--remote-service-deployment-name"})
   private String remoteServiceDeploymentName;
 
+  @CommandLine.Option(names = {"--remote-target-name"})
+  private String remoteTargetName;
+
```
majanjua-amzn, line 80: Not needed

validator/src/main/java/com/amazon/aoc/App.java
```diff
@@ -160,8 +163,9 @@ public Integer call() throws Exception {
     context.setServiceName(this.serviceName);
     context.setRemoteServiceName(this.remoteServiceName);
     context.setRemoteServiceDeploymentName(this.remoteServiceDeploymentName);
+    context.setRemoteTargetName(this.remoteTargetName);
```
majanjua-amzn, line 166: Not needed

validator/src/main/java/com/amazon/aoc/models/Context.java
```diff
@@ -49,9 +49,11 @@ public class Context {
 
   private String remoteServiceDeploymentName;
 
+  private String remoteTargetName;
+
```
majanjua-amzn, line 53: Not needed

validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
```diff
@@ -90,29 +92,47 @@ public void validate() throws Exception {
...
         () -> {
-          // We will query both the Service and RemoteService dimensions to ensure we get all
-          // metrics from all aggregations, specifically the [RemoteService] aggregation.
+          // We will query metrics that include Service, RemoteService, or RemoteTarget dimensions
+          // to ensure we get all metrics from all aggregations, specifically the [RemoteService] aggregation.
           List<String> serviceNames =
               Lists.newArrayList(
                   context.getServiceName(), context.getRemoteServiceDeploymentName());
-          // TODO - Put the following back in: "www.amazon.com",  "AWS.SDK.S3"
+          // TODO - Put the following back in: "www.amazon.com"
```
majanjua-amzn, line 100: I don't think we will put this back, so you can remove the comment entirely

validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
```diff
@@ -90,29 +92,47 @@ public void validate() throws Exception {
...
+                    new Pair<>(CloudWatchService.REMOTE_SERVICE_DIMENSION, remoteServiceName)));
+          }
+          // Query for metrics that includes both of these <remoteService, remoteTarget> values.
+          // Querying just 'remoteService="AWS.SDK.S3"' would also work, but that can result in
+          // returning too many canary metrics and may cause issues.
+          if (context.getTestingId() != null && !context.getTestingId().isEmpty()) {
+            dimensionLists.add(
+                Arrays.asList(
+                    new Pair<>(CloudWatchService.REMOTE_SERVICE_DIMENSION, "AWS.SDK.S3"),
+                    new Pair<>(CloudWatchService.REMOTE_TARGET_DIMENSION, "::s3:::e2e-test-bucket-name-" + context.getTestingId())));
```
majanjua-amzn, line 129: Hardcoding `AWS.SDK.S3` and `::s3:::e2e-test-bucket-name-` is not great. This value should be pulled from the expected data templates if possible. I can see how that would be difficult though so this should be fine.

validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
```diff
@@ -90,29 +92,47 @@ public void validate() throws Exception {
...
+                Arrays.asList(
+                    new Pair<>(CloudWatchService.REMOTE_SERVICE_DIMENSION, remoteServiceName)));
+          }
+          // Query for metrics that includes both of these <remoteService, remoteTarget> values.
+          // Querying just 'remoteService="AWS.SDK.S3"' would also work, but that can result in
+          // returning too many canary metrics and may cause issues.
+          if (context.getTestingId() != null && !context.getTestingId().isEmpty()) {
+            dimensionLists.add(
+                Arrays.asList(
+                    new Pair<>(CloudWatchService.REMOTE_SERVICE_DIMENSION, "AWS.SDK.S3"),
```
majanjua-amzn, line 128: Wouldn't looking for `AWS.SDK.S3` still result in many unwanted metrics? Are we querying that value alongside the bucket name to ensure we don't get unwanted metrics? Just making sure my understanding is correct.

.github/workflows/appsignals-e2e-eks-test.yml
jj22ee: For `aws-otel-java-instrumentation`, this workflow isn't used, the java repo uses its [own copy of this workflow](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/workflows/appsignals-e2e-eks-test.yml#L14). In this repository on the other hand, this isn't used anymore. To be extra safe, I'll undo this change, keep it as is until we remove it entirely.

validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
jj22ee: Since values are usually provided to the templates, not pulled from them, another alternative to remove hard-coded values is to [undo this previous change](https://github.com/aws-observability/aws-application-signals-test-framework/pull/41/commits/d21a9e07ef76e3627396458cdee0dc4d796d6f14) and keep usage of `--remote-target-name`, and apply similar logic for `AWS.SDK.S3`. Unless we we want that logic back, I'll leave this as is.

validator/src/main/java/com/amazon/aoc/App.java
jj22ee: Removed

validator/src/main/java/com/amazon/aoc/App.java
jj22ee: Removed

validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
jj22ee: Removed

validator/src/main/java/com/amazon/aoc/models/Context.java
jj22ee: Removed

validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
majanjua-amzn: No, I think this is fine for now since we are looking for very specific values. Is looking for `RemoteService == AWS.SDK.S3` even necessary though, given we will also be looking for the `RemoteTarget` which is a unique value?

validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
majanjua-amzn: Yeah, then as I stated in the reply to the comment thread above this one the `RemoteService == AWS.SDK.S3` is not needed anyways since we're looking for `bucket-name` which is unique.

Perhaps this can help make the code less complicated. Your refactor to use a list of pairs of dimension key-values works well but if we can avoid having such complicated logic it would be best.

jj22ee: Removed refactor, now using previous logic to query `bucket-name`.

PR #43, opened 2024-04-17
Update Gradlew Daemon to use existing build in cache

harrryr:
*Issue #, if available:*
When the gradlew daemon restarts it needs to download dependency to build the .jar files. This can cause transient issues if the download fails

*Description of changes:*
Make the daemon use cached build from ~/.gradle/build instead
We know this is working because if you look at the test [run](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8717247804/job/23912139825#step:13:29), it shows `FROM-CACHE` at the end of each build.

Test run: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8717247804
Test run after comment: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8727081926
Test run after adding retries and wait time: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8730663803/job/23954844966

### DIFF
--- a/.github/workflows/appsignals-e2e-ec2-test.yml
+++ b/.github/workflows/appsignals-e2e-ec2-test.yml
@@ -73,8 +73,8 @@ jobs:
       - name: Deploy sample app via terraform and wait for endpoint to come online
         working-directory: terraform/ec2
         run: |
-          # Attempt to deploy the sample app on an EC2 instance and wait for its endpoint to come online. 
-          # There may be occasional failures due to transitivity issues, so try up to 2 times. 
+          # Attempt to deploy the sample app on an EC2 instance and wait for its endpoint to come online.
+          # There may be occasional failures due to transitivity issues, so try up to 2 times.
           # deployment_failed of 0 indicates that both the terraform deployment and the endpoint are running, while 1 indicates
           # that it failed at some point
           retry_counter=0
@@ -90,11 +90,11 @@ jobs:
               -var="get_cw_agent_rpm_command=${{ env.GET_CW_AGENT_RPM_COMMAND }}" \
               -var="get_adot_jar_command=${{ env.GET_ADOT_JAR_COMMAND }}" \
             || deployment_failed=$?
-          
+
             if [ $deployment_failed -eq 1 ]; then
               echo "Terraform deployment was unsuccessful. Will attempt to retry deployment."
             fi
-          
+
             # If the deployment_failed is still 0, then the terraform deployment succeeded and now try to connect to the endpoint.
             # Attempts to connect will be made for up to 10 minutes
             if [ $deployment_failed -eq 0 ]; then
@@ -108,12 +108,12 @@ jobs:
                   deployment_failed=1
                   break
                 fi
-          
+
                 printf '.'
                 attempt_counter=$(($attempt_counter+1))
                 sleep 10
               done
-          
+
               echo "Attempting to connect to the remote sample app endpoint"
               remote_sample_app_endpoint=http://$(terraform output sample_app_remote_service_public_ip):8080/healthcheck
               attempt_counter=0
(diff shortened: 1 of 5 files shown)


.github/workflows/appsignals-e2e-ec2-test.yml
```diff
@@ -171,6 +171,9 @@ jobs:
           curl -S -s http://${{ env.MAIN_SERVICE_ENDPOINT }}/remote-service?ip=${{ env.REMOTE_SERVICE_IP }}/
           curl -S -s http://${{ env.MAIN_SERVICE_ENDPOINT }}/client-call/
 
+      - name: Initiate Gradlew Daemon
+        run: ./gradlew
```
majanjua-amzn, line 175: Regardless of the improvement, can we wrap this step in a retry?

harrryr: Done

Dockerfile
```diff
@@ -40,9 +40,19 @@ RUN mv /tmp/eksctl /usr/local/bin
...
 COPY gradlew.bat .
+COPY gradle.properties .
 COPY gradle/ /gradle/
 COPY buildSrc/ /buildSrc/
 COPY validator/ /validator/
 
 # Build gradlew here so that the canary doesn't spend time downloading and building the package
-RUN ./gradlew
\ No newline at end of file
+RUN ./gradlew
```
zzhlogin, line 49: We run `./gradlew` twice in both Dockerfile and workflow. Is it possible to run only once?

harrryr: No unfortunately when we start the container in the workflow the Docker Daemon is in a stopped state, so we have to run it again to initialize it.

.github/workflows/appsignals-e2e-eks-test.yml
```diff
@@ -258,6 +258,12 @@ jobs:
           curl -S -s http://${{ env.APP_ENDPOINT }}/remote-service?ip=${{ env.REMOTE_SERVICE_POD_IP }}/
           curl -S -s http://${{ env.APP_ENDPOINT }}/client-call/
 
+      - name: Initiate Gradlew Daemon
+        uses: ./.github/workflows/actions/execute_and_retry
+        with:
+          command: "./gradlew"
+          cleanup: "./gradlew clean"
```
majanjua-amzn, line 265: We can set the number of retries here higher

majanjua-amzn: Also, it doesn't look like the gradle is picking up the fact that the daemon is there [link](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8727081926/job/23943790575#step:13:82). Is building gradle in advance still helping somehow?

harrryr: Yeah building gradle in advance lets us reuse the cache so that we don't have to download it again. 
Gradle is picking up that the daemon is there, it's just shutdown so we have to start a new one

Dockerfile
zzhlogin: What if we only run `./gradlew` in workflow without in Dockerfile?

harrryr: The point of running the ./gradlew in Dockerfile is so that we don't have to download the dependencies during our canary run. If we run it in the canary, sometimes it may fail due to transient issues and thereby fail the e2e test. We are trying to migrate as much as possible to the DockerFile to mitigate this issue.

harrryr: By running it in the Dockerfile, we have the build cache saved in our docker image, which we can reuse when we call ./gradlew in the E2E testing.

zzhlogin (approved):

zzhlogin (approved):

PR #56, opened 2024-04-29
Remove amazon-cloudwatch delete and call APIs before remote sample app

harrryr:
*Issue #, if available:*
Metric limiter is having frequent failures due to two issues:
1. remote-sample-app `/healthcheck` is called before the sample-app generates two APIs, causing the metric to have the actual value instead of `AllOtherRemoteOperation` ([Sample Link](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8885563298/job/24397189617))
2. during amazon-cloudwatch delete namespace, the amazon-cloudwatch-observability-controller doesn't restart properly and ADOT isn't installed. ([Sample Link](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8874189740/job/24361212097#step:14:608))

*Description of changes:*
1. Call two sample app APIs before pinging the remote sample app
2. Do not delete amazon-cloudwatch. It is not a necessary operation

Test run: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/8886143322


majanjua-amzn (approved):
Approved, but I find the removal of the delete/wait lines concerning. Why would deleting the pods cause ADOT not to be installed properly?

harrryr:
I believe `amazon-cloudwatch-observability-controller` is responsible for installing ADOT onto the sample applications. I'm not sure why it sometimes doesn't restart after deleting though.

PR #65, opened 2024-05-13
Update E2E Tests Expected Data for K8s

jj22ee:
*Issue #, if available:*
Followup from: https://github.com/aws-observability/aws-application-signals-test-framework/pull/41
Previously, new metric rollup test cases were added for EC2 and EKS e2e tests in Python and Java. These need to be added for K8s e2e tests as well.

*Description of changes:*
Updating k8s tests (workflow and expected data) to have the same changes [previously applied for EC2/EKS](https://github.com/aws-observability/aws-application-signals-test-framework/commit/1690fde15c8a5a5889a5f97bc1bb51453b47b85f).

Additionally, changes were made to make the generated data between logs/metrics/traces to be more consistent in the tests.

*Testing:*
Ran K8s tests a few times In IAD using a node for testing changes: https://github.com/aws-observability/aws-application-signals-test-framework/actions/workflows/test.yml
- https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/9069465891

### DIFF
--- a/.github/workflows/appsignals-e2e-k8s-test.yml
+++ b/.github/workflows/appsignals-e2e-k8s-test.yml
@@ -101,10 +101,10 @@ jobs:
       - name: Call all test APIs
         continue-on-error: true
         run: |
-          curl -S -s http://${{ env.MAIN_SERVICE_ENDPOINT }}:30100/outgoing-http-call/; echo
-          curl -S -s http://${{ env.MAIN_SERVICE_ENDPOINT }}:30100/aws-sdk-call/; echo
-          curl -S -s http://${{ env.MAIN_SERVICE_ENDPOINT }}:30100/remote-service?ip=${{ env.REMOTE_SERVICE_IP }}/; echo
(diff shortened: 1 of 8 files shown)


majanjua-amzn (commented):
There should be changes in `aws-sdk-call-log.mustache` and `aws-sdk-call-trace.mustache` as well. Those files use RegEx so the difference in value wasn't detected because the RegEx `::s3:::e2e-test-bucket-name` still detects `::s3:::e2e-test-bucket-name-{{testingId}}`

You should replace that RegEx with `^::s3:::e2e-test-bucket-name-{{testingId}}$` which enforces the start and end of the string.

jj22ee:
> There should be changes in aws-sdk-call-log.mustache and aws-sdk-call-trace.mustache as well

For trace and logs tests, the workflows do not send the Testing ID to the Sample App. So for these logs/traces tests, the S3 bucket name will still be `::s3:::e2e-test-bucket-name`

majanjua-amzn (approved):

PR #73, opened 2024-05-29
Update Metric Limiter Validation

harrryr:
*Issue #, if available:*
Currently the metric limiter test is failing too frequently due to issues with the API requests. Since the ultimate goal of this test is to ensure that the operation value `AllOtherOperations` show up properly, we are relaxing the requirements for the validation by only checking if the value shows up for any of the four test APIs. 

Test run: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/9289110511


validator/src/main/resources/validations/metric_limiter/metric-validation.yml
```diff
@@ -3,22 +3,4 @@
   httpPath: "/outgoing-http-call"
   httpMethod: "get"
   callingType: "http"
```
mxiamxia, line 5: it looks we expect `AllOtherOperation` only from `httpPath: "/outgoing-http-call"` API? Will it possible that `/outgoing-http-call` will not be aggregated by MetricLimiter? should we check `AllOtherOperation` on all 4 APIs if any of it being populated with ``AllOtherOperation``?

mxiamxia: I am worried if `/outgoing-http-call` is processed normally without MetricLimiter involved, the MetricLimiter applies for the rest of APIs

harrryr: We do check on all 4 APIs. We call all four APIs here: https://github.com/aws-observability/aws-application-signals-test-framework/blob/main/.github/workflows/appsignals-e2e-metric-limiter-test.yml#L282-L285

Even if the `outgoing-http-call` is not aggregated by the Metric Limiter, since we called the other APIs in the link above, the metric validator should still be able to find the `AllOtherOperation` metric

The reason why I added` /outgoing-http-call` in the validation is because the metric validator is designed to look for the {traceid: "1-xxxx-xxxx"} format once it makes a call. I tried using `/` instead but it throws an error because the returned value is healthcheck, not traceid.

harrryr: > I am worried if `/outgoing-http-call` is processed normally without MetricLimiter involved, the MetricLimiter logic applies to the rest of APIs instead

Yes, even if that happens the metric validator should still pass because the it's looking for a schema with [Service, Operation, HostedIn.EKS.Cluster, HostedIn.K8s.Namespace]. Since all four APIs generate the same value for this schema, if this schema can be found with operation = AllOtherOperation, we'll know that the metric aggregation occurred.

validator/src/main/resources/expected-data-template/metric_limiter/metric-limiter-metric.mustache
```diff
@@ -0,0 +1,92 @@
+-
+  metricName: Latency
+  namespace: {{metricNamespace}}
+  dimensions:
+    -
+      name: Operation
+      value: AllOtherOperations
```
bjrara, line 7: Do we still have logics to verify any operation escaped from metrics limiting? The threshold we set is still 2, right?

harrryr: The threshold is still 2. We still send two irrelevant API calls to the sample app initially to make it so that we exceed the threshold.

mxiamxia (approved):
thanks!

bjrara (approved):
I don't think it's a blocking one, but it would be nice to have some logics to check if we have metrics that are not rolled up with the original `Operation` values in place.

PR #89, opened 2024-06-11
Check if Python Remote App Endpoint is Up

harrryr:
*Issue #, if available:*
Occasionally the remote sample app for Python doesn't initiate properly and causes errors during metric validation

*Description of changes:*
Add a step to check if Python remote sample app endpoint is properly initiated.

Test run: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/9487123419

### DIFF
--- a/terraform/python/eks/main.tf
+++ b/terraform/python/eks/main.tf
@@ -263,6 +263,45 @@ resource "kubernetes_service" "python_r_app_service" {
   }
 }
 
+resource "kubernetes_ingress_v1" "python-r-app-ingress" {
+  depends_on = [kubernetes_service.python_r_app_service]
+  wait_for_load_balancer = true
+  metadata {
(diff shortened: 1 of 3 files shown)


terraform/python/eks/main.tf
```diff
@@ -263,6 +263,45 @@ resource "kubernetes_service" "python_r_app_service" {
   }
 }
 
+resource "kubernetes_ingress_v1" "python-r-app-ingress" {
```
majanjua-amzn, line 266: Why did we not have this before? Is this needed? Do we do the same for Java?

harrryr: It's only needed to check if the remote sample application endpoint is available. Yes, we do the same for Java

majanjua-amzn (approved):

PR #90, opened 2024-06-12
Increase Retry Wait Times and Add Additional Gradlew Step

harrryr:
*Issue #, if available:*
Increasing the wait times for `Terraform Init`, and `Gradlew` Step to mitigate the occurrence of these issues failing, especially `Terraform Init`. 

For Gradlew, wait time is 60 and retry is 3. If it fails initially, will try again with the same parameters. If retry triggers, it will add a max of 6 minutes. 

For Terraform Init, wait time is 60 and retry is 6. If retry triggers, it will add a max of 6 minutes

Most runs are 8-10 minutes without the retries, so these additional retries could cause the runs to last more than 15 minutes. Since it's rare for these errors to occur consecutively, should not be an issue if it exceeds 15 minutes by a minute or two. 

Additionally, adding `Initiate Gradlew Step` to the beginning of the run so that if it fails, we can try it again before the validator step. Cannot do the same for `Terraform Init` because this steps already occurs very early on during the run. 

The task also involved increasing wait time for the enable-app-signal.sh script, but since the cleanup command wasn't working and is fixed on this [PR](https://github.com/aws-observability/aws-application-signals-test-framework/pull/88), will leave it as it is for now. 

Test run: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/9489300435
Test run after adding continue-on error: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/9508162874

### DIFF
--- a/.github/workflows/application-signals-java-e2e-k8s-test.yml
+++ b/.github/workflows/application-signals-java-e2e-k8s-test.yml
@@ -60,7 +60,7 @@ jobs:
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v4
         with:
-          role-to-assume: ${{ secrets.E2E_SECRET_TEST_ROLE_ARN }}
+          role-to-assume: ${{ secrets.E2E_IAD_TEST_ACCOUNT_ARN }}
           aws-region: us-east-1
 
       - name: Retrieve account


majanjua-amzn (approved):

.github/workflows/application-signals-java-e2e-k8s-test.yml
```diff
@@ -120,6 +131,15 @@ jobs:
           curl -S -s "http://${{ env.MAIN_SERVICE_ENDPOINT }}:30100/remote-service?ip=${{ env.REMOTE_SERVICE_IP }}&testingId=${{ env.TESTING_ID }}"; echo
           curl -S -s "http://${{ env.MAIN_SERVICE_ENDPOINT }}:30100/client-call"; echo
 
+      - name: Initiate Gradlew Daemon
+        if: steps.initiate-gradlew == 'failure'
+        uses: ./.github/workflows/actions/execute_and_retry
+        with:
+          command: "./gradlew"
```
majanjua-amzn, line 138: Why does it matter if the first try failed? Does this try to overwrite the previous attempt?

Check for options that would allow `./gradlew` to just stop if the job is already done.

Also, should we `continue-on-error: true` here as well since we try again when validating too?

harrryr: 1. If the first try failed then we want to run it again to make sure `./gradlew` creates a daemon successfully. That's the point of splitting it to the beginning of the run and right before the validation step
2. That's basically what the `if` statement is doing. If it succeeded don't run ./gradlew again, if failed run it again.
3. Makes sense. Will add

majanjua-amzn (approved):

PR #93, opened 2024-06-18
Create PULL_REQUEST_TEMPLATE.md

majanjua-amzn:
Include testing instructions to ensure we don't break any canaries during merges.


.github/PULL_REQUEST_TEMPLATE.md
harrryr, line 1: Mention that these cannot run in parallel and will need to add a needs: [] statement

harrryr (approved):

PR #96, opened 2024-06-26
Propagate RDS database endpoint and credentials to the SampleApp and expose a new API to query the database

jerry-shao:
*Issue description:*
in Application Signals E2E tests, we are not testing database related span today.

*Description of changes:*
This change will introduce two new Github workflow steps for Java E2E canary to call RDS to get the endpoint and cluster identifier via CLI, store them as Github environment variables, and then use the cluster identifier to call SecretsManager via CLI to get the secrets name. In the second workflow step, it uses secrets manager action to retrieve credentials based on the secrets name for RDS cluster obtained from the previous step.

Java Sample APP is also updated to expose a new API: `mysql` which makes a simple query: `select * from tables limit 1;` from the `information_schema` database which is a informational built-in database for all MySQL databases;

Invoking `/mysql` API and asserting desired span information will be implemented in a separate PR. Same for Python.

This has been tested here: https://github.com/jerry-shao/aws-application-signals-test-framework/actions/runs/9677471098/job/26701733590

### DIFF
--- a/.github/workflows/application-signals-java-e2e-ec2-canary-test.yml
+++ b/.github/workflows/application-signals-java-e2e-ec2-canary-test.yml
@@ -7,8 +7,8 @@
 ## including logs, metrics, and traces.
 name: Application Signals Enablement - Java E2E EC2 Canary Testing
 on:
-#  schedule:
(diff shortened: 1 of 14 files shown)


vastin (approved):
LGTM

PR #99, opened 2024-06-28
feat: Adding unit test for TraceValidator

jerry-shao:
*Issue description:*

There is only unit testing for MetricValidator only, it's hard to understand whether TraceValidator is working as expected.

*Description of changes:*

This change introduces a unit test for TraceValidator which could make sure the TraceValidator is working as expected. Also this could help developers debug the difference between actual traces and the mustache template.

This unit test will also be executed for every PR to make sure new changes won't break validator behaviour.

E2E testing workflow: https://github.com/jerry-shao/aws-application-signals-test-framework/actions/runs/9716864880/job/26821375468
(Note: Only eu-west-1 passed because that's the region I have E2E infrastructure deployed)

### DIFF
--- a/validator/src/main/java/com/amazon/aoc/validators/ValidatorFactory.java
+++ b/validator/src/main/java/com/amazon/aoc/validators/ValidatorFactory.java
@@ -22,6 +22,7 @@ import com.amazon.aoc.exception.ExceptionCode;
 import com.amazon.aoc.fileconfigs.FileConfig;
 import com.amazon.aoc.models.Context;
 import com.amazon.aoc.models.ValidationConfig;
+import com.amazon.aoc.services.XRayService;
 
 public class ValidatorFactory {
   private Context context;
@@ -43,7 +44,7 @@ public class ValidatorFactory {
     FileConfig expectedData = null;
     switch (validationConfig.getValidationType()) {
       case "trace":
-        validator = new TraceValidator();
+        validator = new TraceValidator(new XRayService(context.getRegion()), 2, 5);
         expectedData = validationConfig.getExpectedTraceTemplate();
         break;
       case "cw-metric":
--- a/validator/src/test/test-resources/trace/actual/example-trace-document.json
+++ b/validator/src/test/test-resources/trace/actual/example-trace-document.json
@@ -3,7 +3,7 @@
   "name": "serviceName",
   "http": {
     "request": {
-      "url": "http://127.0.0.1/aws-sdk-call?ip=1.2.3.4&testingId=abc",
+      "url": "http://127.0.0.1/aws-sdk-call?ip=1.2.3.4&testingId=testIdentifier",
       "method": "GET"
     },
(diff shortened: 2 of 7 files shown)


validator/src/main/java/com/amazon/aoc/validators/ValidatorFactory.java
```diff
@@ -43,7 +44,7 @@ public IValidator launchValidator(ValidationConfig validationConfig) throws Exce
     FileConfig expectedData = null;
     switch (validationConfig.getValidationType()) {
       case "trace":
-        validator = new TraceValidator();
+        validator = new TraceValidator(new XRayService(context.getRegion()), 2, 5);
```
majanjua-amzn, line 47: nit: Curious why we would move this out of the `init` function and into the constructor. Other two validators have the default values as `DEFAULT_MAX_RETRY_COUNT` like [here](https://github.com/aws-observability/aws-application-signals-test-framework/blob/17c094f770bab2bcb74f858264adfa75b164e3f9/validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java#L40). Would be best to keep it consistent if possible

validator/src/test/test-resources/trace/actual/example-trace-document.json
majanjua-amzn, line 1: Are there any attributes you removed from the trace or is this as close as possible to a real one with fake values? Just curious

majanjua-amzn (commented):
Great change, just some small questions

validator/src/main/java/com/amazon/aoc/validators/ValidatorFactory.java
jerry-shao: This small refactoring is mainly for unit test. 

Previously the corresponding CloudWatch service(metric, log, trace) is initialised directly within the init method instead of passing it through a parameter which made the unit test hard to mock this service. Hence, these validators are using a setter method for the CloudWatch service just for unit test to be able to overwrite it with a mock. 

Creating a new object in a class is not the best approach as it not just hard to test, but also introduce high coupling. 

My plan is to refactor MetricValidator and LogValidator in the next few PRs to also decouple CloudWatch service from it and get rid of these setter methods.

Same for the retry count, they are defined within the validator. Passing it through constructor could let unit test fail fast instead of retrying til the max limit.

validator/src/test/test-resources/trace/actual/example-trace-document.json
jerry-shao: Good question.

Firstly, I took the s3-sdk-call mustache template as an expected template. Then I found a real corresponding XRay trace for it in my dev account as an actual trace document. I also removed some unnecessary attributes like fault, error, throttle, instance size, start_time, end_time etc. as the main goal of this unit test is to make sure the TraceValidator can compare data model correctly between the template and actual trace.

majanjua-amzn (approved):

PR #105, opened 2024-07-03
Add Terraform Init to Dockerfile and also create ECR per region

harrryr:
*Issue description:*
E2E tests pull image from ECR when workflow starts. Due to the high number of E2E tests using the same ECR, the ECR often throttles and causes the test to fail relatively frequently. This issue will become worse as we add more language and platform support, and currently requires asking for a service quota increase manually. Additionally, another top issue for E2E test failure is due to Terraform failing to initialize.

*Description of changes:*
We decided to allocate an ECR for each region. Since there are currently 9 canaries running, it is not possible for it to exceed the 10 request/s API limit so this should resolve the throttling issue

The `e2e-test-docker-image-build.yml` has been seperated into two workflows: `github-runner-image-build.yml` and  `github-runner-image-push.yml`. The reason for this change is due to the following circumstances:
1. We build a docker image for each different canaries (The only difference between the images is the Terraform Init. This won't have an impact on the API throttling issue, but it was compartmentalize the images to each different use case).
2. Each image is 3.5gb in size. Github runner disk size is 12gb so we can't build all the artifacts and push it to ECR. We have to build and delete the file one by one to manage disk space. Note: Out of the 3.5gb, 700mb is from the Terraform Init, so we considered building all the Terraform code in one image since the final size will be 2.8gb + 0.7gb * 8 = 8.4gb. However, this isn't scalable since if we add 2 more languages we will encounter the same issue again. 
5. To loop through each image, Github API was used. However, Github API doesn't support downloading artifacts from current runs, it only allows downloading artifacts from completed runs.
6. Therefore, two workflow files were created. `github-runner-image-build.yml` will run whenever there is a change in Terraform or Dockerfile code and build the images, `github-runner-image-push.yml` will run whenever the previous workflow run is ran, then push the images to the ECR. 

As mentioned above, Terraform Init is now done during docker build. However, since the build and push process takes around an hour, there is an edge case where the workflow files are updated to the latest PR but the docker image isn't updated yet with the latest Terraform. To resolve this 1 hour period, extra steps were added in the workflow file to check if there is a difference in the latest and docker's Terraform file, and if there is, then run the Terraform Init step inside the workflow. 

*Test*
Several test runs:
- Testing to check if  `github-runner-image-build.yml` builds images properly: [Link](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10012887630)
- Testing to check if `github-runner-image-push.yml` pushes images if provided with the correct run_id: [Link](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10012993832)
- The `workflow_run` condition only works in the main branch. Therefore, this condition was tested on forked repo: ([Build Link](https://github.com/harrryr/aws-application-signals-test-framework/actions/runs/9999519743), [Push Link ](https://github.com/harrryr/aws-application-signals-test-framework/actions/runs/9999588939)with correct run_id
- Testing to check if the Terraform Diff logic triggers if there is difference: [Link](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10014011074)
- E2E test to check if it works with the new images: [Link](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10014266147),  [Link for K8s](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10014433713)

Note: Currently the images have been deployed to all the regions. However, we should run this workflow again after this PR is merged to double check that it works as expected


.github/workflows/e2e-test-docker-image-build.yml
```diff
@@ -30,9 +47,10 @@ jobs:
         id: build-image
         env:
           REGISTRY: ${{ steps.login-ecr-public.outputs.registry }}
-          REGISTRY_ALIAS: h6o3z5z9
-          REPOSITORY: aws-application-signals-test-framework-workflow-container
+          REGISTRY_ALIAS: v6t6x3y1
```
majanjua-amzn, line 50: Thought we were going to have a custom alias now?
https://docs.aws.amazon.com/AmazonECR/latest/public/public-registry-settings.html

Dockerfile
```diff
@@ -55,4 +57,12 @@ ENV GRADLE_USER_HOME=/.gradle/
...
\ No newline at end of file
+RUN cp -r ~/.gradle/* $GRADLE_USER_HOME
+
+COPY "$TERRAFORM_DIR" /terraform/
+RUN if echo "$TERRAFORM_DIR" | grep -q "k8s"; then \
+      terraform -chdir=/terraform/deploy init && terraform -chdir=/terraform/deploy validate ; \
+      terraform -chdir=/terraform/cleanup init && terraform -chdir=/terraform/cleanup validate ; \
+    else \
+      terraform -chdir=/terraform init && terraform -chdir=/terraform validate ; \
+    fi
```
majanjua-amzn, line 68: Nice

Dockerfile
```diff
@@ -5,6 +5,8 @@ FROM openjdk:11-jdk
 ENV JAVA_HOME=/usr/local/openjdk-11
 ENV PATH="$JAVA_HOME/bin:${PATH}"
 
+ARG TERRAFORM_DIR
+
```
majanjua-amzn, line 9: Is there some convention on where to put ARGs in Dockerfiles? Perhaps at the top with some very explicit comment on what they mean

.github/workflows/e2e-test-docker-image-build.yml
```diff
@@ -2,6 +2,12 @@ name: Create and publish a Docker image
 
 on:
   workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - 'Dockerfile'
+      - 'terraform/**'
```
majanjua-amzn, line 10: nit: Is there a way to exclude files? Might want to specifically exclude changes to `*.md` and other similar files since some of them might have now or in the future and changes to those are not relevant. I'll leave it up to you to identify if there's room for that here

.github/workflows/e2e-test-docker-image-build.yml
```diff
@@ -10,14 +16,25 @@ permissions:
...
+                  { repo: 'python-k8s', dir: '/terraform/python/k8s' } ]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
 
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v4
         with:
-          role-to-assume: ${{ secrets.E2E_SECRET_TEST_ROLE_ARN }}
+          role-to-assume: ${{ secrets.E2E_IAD_TEST_ACCOUNT_ARN }}
```
majanjua-amzn, line 37: The value changed from the pipeline account to the IAD test account, right?

.github/workflows/e2e-test-docker-image-build.yml
harrryr: This isn't needed anymore. We retrieve the url programmatically

Dockerfile
harrryr: Ty

Dockerfile
harrryr: Added comment

.github/workflows/e2e-test-docker-image-build.yml
harrryr: Hmm, we could add this if the files are added in the future. Not a huge concern because these future presumably will not be edited often and running the workflow to update the images isn't really a concern

.github/workflows/e2e-test-docker-image-build.yml
harrryr: Yes

majanjua-amzn (commented):
Can we split this PR into two PRs? One for image build/push, then we can run a test with the officially created new images, then we can update all the canaries in a second PR

.github/workflows/github-runner-image-build.yml
```diff
@@ -0,0 +1,46 @@
...
+        run: docker build -t ${{ matrix.terraform-dir.name }}:latest --build-arg TERRAFORM_DIR=${{ matrix.terraform-dir.dir }} .
+
+      - name: Save docker image as .tar
+        run: docker save ${{ matrix.terraform-dir.name }}:latest > ${{ matrix.terraform-dir.name }}.tar
+
+      - name: Upload docker image
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.terraform-dir.name }}.tar
+          path: ${{ matrix.terraform-dir.name }}.tar
```
majanjua-amzn, line 46: To where? Not sure I understand what's happening here, where do the images built go if they're not pushed immediately?
Does this rely on GH's version of ECR?

.github/workflows/github-runner-image-push.yml
majanjua-amzn, line 1: How do we enforce the push is run right after the build?

.github/workflows/github-runner-image-push.yml
```diff
@@ -0,0 +1,105 @@
...
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::${{ env.ACCOUNT_ID }}:role/${{ env.E2E_TEST_ROLE_NAME }}
+          aws-region: us-east-1
+
+      - name: Output Public ECR Url
+        id: get-public-ecr-url
+        run: |
+          echo "public_ecr_url=$(aws ecr-public describe-repositories --repository-names ${{ env.E2E_RUNNER_ECR_NAME }} --query "repositories[0].repositoryUri" --output text)" >> $GITHUB_OUTPUT
```
majanjua-amzn, line 55: I assume we can use `[0]` because we're specifically looking for `E2E_RUNNER_ECR_NAME`?

.github/workflows/github-runner-image-push.yml
```diff
@@ -0,0 +1,105 @@
+# This workflow will trigger when it detects that a new github runner image was built in the
+# github-runner-image-build.yml workflow. It will download the artifacts built from that workflow
+# and push it to an ECR in each region
+name: Push Github Runner Image
+
+on:
+  workflow_run:
+    workflows:
+      - "Build Github Runner Image"
```
majanjua-amzn, line 9: So this runs after the other workflow or does it trigger when the other triggers? ie. are they parallelized or serialized?

majanjua-amzn (commented):
Can we do this for one canary first, like Java EKS, to see the result?

.github/workflows/github-runner-image-build.yml
harrryr: Github can store artifacts created by workflows. The artifacts are attached to the workflow run that created them. Not ECR, it can store any files such as .txt, .tar, etc. As you can see, we converted the image to a .tar file and uploaded it

.github/workflows/github-runner-image-push.yml
harrryr: They are serialized. It get triggered when the "Build Github Runner Image" workflow run is completed on the main branch

.github/workflows/github-runner-image-push.yml
harrryr: Yes

.github/workflows/github-runner-image-push.yml
harrryr: Through the workflow run condition `"Build Github Runner Image", type: "completed"`

.github/workflows/java-ec2-asg-e2e-test.yml
```diff
@@ -31,10 +31,39 @@ env:
   TEST_RESOURCES_FOLDER: ${GITHUB_WORKSPACE}
 
 jobs:
+  get-public-ecr-url:
```
majanjua-amzn, line 34: Can we extract this as an action?

PR #115, opened 2024-07-12
Propagate RDS database endpoint and credentials to the SampleApp and expose a new API to query the database for Python

georgeboc:
*Issue description:*
In Application Signals E2E tests for Python, we are not testing database related span today.

*Description of changes:*
This change will introduce two new Github workflow steps for Python E2E canary to call RDS to get the endpoint and cluster identifier via CLI, store them as Github environment variables, and then use the cluster identifier to call SecretsManager via CLI to get the secrets name. In the second workflow step, it uses secrets manager action to retrieve credentials based on the secrets name for RDS cluster obtained from the previous step. This PR is based off [this other PR](https://github.com/aws-observability/aws-application-signals-test-framework/pull/96) for Java.

The Python Sample APP is also updated to expose a new API: mysql which makes a simple query: select * from tables limit 1; from the information_schema database which is a informational built-in database for all MySQL databases;

This has been tested here: https://github.com/georgeboc/aws-application-signals-test-framework/actions/runs/9905604317/job/27365487080

We can safely merge this PR even the RDS database is not available as we are not using it yet. The workflow continues on these steps when we get error from RDS saying the cluster was not found.


harrryr (approved):

.github/workflows/application-signals-python-e2e-eks-test.yml
```diff
@@ -157,6 +174,10 @@ jobs:
               -var="service_account_aws_access=service-account-${{ env.TESTING_ID }}" \
               -var="python_app_image=${{ env.ACCOUNT_ID }}.dkr.ecr.${{ inputs.aws-region }}.amazonaws.com/${{ secrets.APP_SIGNALS_PYTHON_E2E_FE_SA_IMG }}" \
               -var="python_remote_app_image=${{ env.ACCOUNT_ID }}.dkr.ecr.${{ inputs.aws-region }}.amazonaws.com/${{ secrets.APP_SIGNALS_PYTHON_E2E_RE_SA_IMG }}" \
+              -var="rds_mysql_cluster_endpoint=${{env.RDS_MYSQL_CLUSTER_ENDPOINT}}" \
+              -var="rds_mysql_cluster_username=${{env.RDS_MYSQL_CLUSTER_SECRETS_USERNAME}}" \
+              -var="rds_mysql_cluster_password=${{env.RDS_MYSQL_CLUSTER_SECRETS_PASSWORD}}" \
+              -var="rds_mysql_cluster_database=information_schema" \
```
majanjua-amzn, line 180: Will the EC2 test start to fail if we don't provide the same environment variables?
Why did we not add these for those test cases as well?

sample-apps/python/django_frontend_service/requirements.txt
majanjua-amzn, line 1: FYI: EC2 test cases use `ec2-requirements.txt`.

.github/workflows/application-signals-python-e2e-eks-test.yml
georgeboc: All those values have defaults, this PR is pretty similar to the [Java PR here](https://github.com/aws-observability/aws-application-signals-test-framework/pull/96), where we also added defaults in the `variables.tf` file.

sample-apps/python/django_frontend_service/frontend_service_app/views.py
```diff
@@ -105,3 +107,26 @@ def get_xray_trace_id():
...
+        cur.execute("SELECT * FROM tables LIMIT 1;")
+        return get_xray_trace_id()
+    except Exception as exception:  # pylint: disable=broad-except
+        logger.exception("Exception Occurred")
+        raise exception
+    finally:
+        if cur:
+            cur.close()
+        if conn:
+            conn.close()
```
jerry-shao, line 132: Nitpick: Could we use Python `with-resource-open` feature to let it close connections automatically for us?
See: https://github.com/PyMySQL/PyMySQL

sample-apps/python/django_frontend_service/requirements.txt
georgeboc: EC2 end-to-end tests won't be in-scope for this PR, only EKS.

sample-apps/python/django_frontend_service/frontend_service_app/views.py
georgeboc: Yes, will address this in the next commit

sample-apps/python/django_frontend_service/requirements.txt
georgeboc: Actually, I tested the e2e tests for the ec2 and k8s platforms in my own fork and they failed because the pymysql dependency was missing. I have changed the `ec2-requirements.txt` file too in the last commits.

sample-apps/python/django_frontend_service/frontend_service_app/views.py
```diff
@@ -105,3 +107,18 @@ def get_xray_trace_id():
...
+                                     user=os.environ["RDS_MYSQL_CLUSTER_USERNAME"],
+                                     password=os.environ["RDS_MYSQL_CLUSTER_PASSWORD"],
+                                     database=os.environ["RDS_MYSQL_CLUSTER_DATABASE"])
+        with connection:
+            with connection.cursor() as cursor:
+                cursor.execute("SELECT * FROM tables LIMIT 1;")
+                return get_xray_trace_id()
+    except Exception as exception:  # pylint: disable=broad-except
+        logger.exception("Exception Occurred")
+        raise exception
```
majanjua-amzn, line 124: For our other APIs, we `return get_xray_trace_id()` regardless of the outcome of the API call, as we don't assume the trace ID was not generated despite an exception being thrown. It's possible that despite an exception a trace ID is still generated and it's important for us to be able to see if the trace generated by the instrumentation on the application is also appearing in the console.
```suggestion
            with connection.cursor() as cursor:
                cursor.execute("SELECT * FROM tables LIMIT 1;")
                return get_xray_trace_id()
    except Exception as exception:  # pylint: disable=broad-except
        logger.error("Could not complete http request to RDS database:" + str(e))
        return get_xray_trace_id()
```

.github/workflows/python-eks-e2e-test.yml
```diff
@@ -174,6 +191,10 @@ jobs:
               -var="service_account_aws_access=service-account-${{ env.TESTING_ID }}" \
               -var="python_app_image=${{ env.MAIN_SAMPLE_APP_IMAGE_ARN }}" \
               -var="python_remote_app_image=${{ env.REMOTE_SAMPLE_APP_IMAGE_ARN }}" \
+              -var="rds_mysql_cluster_endpoint=${{env.RDS_MYSQL_CLUSTER_ENDPOINT}}" \
+              -var="rds_mysql_cluster_username=${{env.RDS_MYSQL_CLUSTER_SECRETS_USERNAME}}" \
+              -var="rds_mysql_cluster_password=${{env.RDS_MYSQL_CLUSTER_SECRETS_PASSWORD}}" \
+              -var="rds_mysql_cluster_database=information_schema" \
```
majanjua-amzn, line 197: Just confirming, this won't cause a failure if we don't deploy the changes to the sample app right? From my understanding it will just add environment variables that won't be used if we don't update the sample app?

georgeboc: That's right, the sample app won't fail if we don't update the version in ECR straightaway. What could happen is that in the GitHub workflow step that performs the curl, will return Error 404 page not found when trying to call the mysql endpoint, although it won't fail the test run anyway because of [this configuration](https://github.com/aws-observability/aws-application-signals-test-framework/blob/main/.github/workflows/python-eks-e2e-test.yml#L309).

sample-apps/python/django_frontend_service/frontend_service_app/views.py
georgeboc: Addressed this comment in the last commits.

majanjua-amzn (approved):

PR #116, opened 2024-07-12
Ec2 cleanup workflow and script

srprash:
### Description of changes:
- Create a workflow that runs every day at 18:00 UTC (11AM PST) to perform clean up of AWS resources in the E2E testing accounts in all regions. The workflow can also be run on demand if needed.
- For now the workflow only cleans up the EC2 resources that are:
  - more than 3 hrs old at the time the workflow runs
  - do NOT belong to an EKS cluster (by checking their security group name)
  - do NOT belong to a K8s cluster (by checking the key name)
  - do NOT have the `do-not-delete` tag.
- It is possible that we may unintentionally terminate some hosts. The cleanup job will upload the full report of instances filtered for termination and upload to an S3 bucket so that we can audit them and fix/improve our filter logic.

NOTE: I will submit another PR to uncomment the termination logic.

### Testing
- Test run: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/9913944002/job/27392017400
  - The bucket name was intentionally incorrect since it is a WIP.
  - The termination part was commented out intentionally. We will process terminations once this PR is reviewed and merged.  

--- 


.github/workflows/resource-cleanup.yml
```diff
@@ -0,0 +1,55 @@
...
+        aws-region: ['af-south-1','ap-east-1','ap-northeast-1','ap-northeast-2','ap-northeast-3','ap-south-1','ap-south-2','ap-southeast-1',
+                     'ap-southeast-2','ap-southeast-3','ap-southeast-4','ca-central-1','eu-central-1','eu-central-2','eu-north-1',
+                     'eu-south-1','eu-south-2','eu-west-1','eu-west-2','eu-west-3','il-central-1','me-central-1','me-south-1', 'sa-east-1',
+                     'us-east-1','us-east-2', 'us-west-1', 'us-west-2']
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - uses: actions/checkout@v3
```
majanjua-amzn, line 25: Does this automatically checkout the current branch or does it default to `main`?

tool/clean/ec2_instance_cleanup/cleaner.py
majanjua-amzn, line 1: Can we add this to `.github/workflows/util` instead? Perhaps still under a `clean` folder in there. I think it fits there and the folder already exists

tool/clean/ec2_instance_cleanup/cleaner.py
```diff
@@ -0,0 +1,139 @@
...
+import logging
+
+import boto3
+from _datetime import datetime, timedelta, timezone
+from typing import List
+
+INSTANCE_STATE_RUNNING = 'running'
+EKS_CLUSTER_SECURITY_GROUP_PREFIX = 'eks-cluster-sg-'
+K8S_INSTANCE_NAME_PREFIX = 'k8s-'
+K8S_KEYNAME_PREFIX = 'k8s'
```
majanjua-amzn, line 12: nit: Is there a reason we can't do this? Does `KEYNAME` not start with a `-` as well?
```suggestion
K8S_RESOURCE_NAME_PREFIX = 'k8s-'
```

majanjua-amzn (approved):

.github/workflows/resource-cleanup.yml
srprash: As per the docs: https://github.com/actions/checkout

>  The branch, tag or SHA to checkout. When checking out the repository that triggered a workflow, this defaults to the reference or SHA for that event. Otherwise, uses the default branch.

So in this case this will the `main` branch.

tool/clean/ec2_instance_cleanup/cleaner.py
srprash: Yea, sure. I think it makes sense. Will move it there.

tool/clean/ec2_instance_cleanup/cleaner.py
srprash: Did a quick spotcheck and yes, the KEYNAMEs start with `k8s-`. I will update this.

.github/workflows/util/clean/ec2_instance_cleanup/cleaner.py
```diff
@@ -0,0 +1,139 @@
...
+import logging
+
+import boto3
+from _datetime import datetime, timedelta, timezone
+from typing import List
+
+INSTANCE_STATE_RUNNING = 'running'
+EKS_CLUSTER_SECURITY_GROUP_PREFIX = 'eks-cluster-sg-'
+K8S_INSTANCE_NAME_PREFIX = 'k8s-'
+K8S_KEYNAME_PREFIX = 'k8s-'
```
majanjua-amzn, line 12: nit: We can merge these into one
```suggestion
K8S_RESOURCE_NAME_PREFIX = 'k8s-'
```

srprash: I think I would like to keep them separate since they are part of two different checks even though they have same value (for now).

majanjua-amzn (approved):

PR #117, opened 2024-07-16
Consolidate release final

harrryr:
*Issue description:*
Separating this [PR](https://github.com/aws-observability/aws-application-signals-test-framework/pull/109) into two to make it easier for review: 

Test Run: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/9961752583

### DIFF
--- a/.github/workflows/java-ec2-e2e-canary-test.yml
+++ b/.github/workflows/java-ec2-e2e-canary-test.yml
@@ -16,7 +16,7 @@ permissions:
   contents: read
 
 jobs:
-  e2e-test-1:
+  default:
     strategy:
       fail-fast: false
       matrix:
@@ -30,7 +30,7 @@ jobs:
       aws-region: ${{ matrix.aws-region }}
       caller-workflow-name: 'appsignals-e2e-ec2-canary-test'
 
-  e2e-test-2:
+  asg:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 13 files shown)


.github/workflows/java-ec2-e2e-canary-test.yml
```diff
@@ -16,29 +16,29 @@ permissions:
   contents: read
 
 jobs:
-  e2e-ec2-default-test:
+  e2e-test-1:
```
majanjua-amzn, line 19: nit: Should still reference the use case right? How does this appear in the workflow later? Your `test.yml` named it properly I think so we should have this be correct too and in line with the rest. Same applies to ASG case and to python if it's named the same way

harrryr: We decided to not reference the jobs with the use case and make it just e2e-test since it's clear what canary is running based on the workflow title and prioritization went to reducing the namespace as much as possible. 

I guess in this case though e2e-test-1 isn't descriptive enough to determine whether default or asg case is running. 

I will update it to e2e-test-asg and e2e-test-default for ec2 use cases then. 

Or what if we just split the canary workflow file into two? One for asg and one for default. That way we don't clog the actions dashboard with 64 regions.

majanjua-amzn: I think the way it is is fine because it covers one language/platform combination, ie. java-ec2 and python-ec2.

Renaming to `asg` and `default` is even enough, the engineer looking at the test can see from the title of the workflow it's for a specific language. Your proposed names `e2e-test-asg` and `e2e-test-default` also work.

You can see in [this run of java/eks](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/9965148669) that `e2e-test (region)` appears as the name of the job. I think `asg (region)` and `default (region)` is sufficient.

Technically the others should be named the same way too but they only have one use case per workflow run so it doesn't matter at all.

harrryr: Updated for all canary workflows
- metric-limiter
- asg
- default
- eks

majanjua-amzn (approved):

PR #149, opened 2024-07-30
Remove docker runner container image

harrryr:
*Issue description:*
To resolve issues where downloading dependencies caused failures in E2E workflows due to transient issues, we began running E2E workflows in a docker image container with those dependencies already installed in this [PR](https://github.com/aws-observability/aws-application-signals-test-framework/pull/33). 

As we started scaling our canary to encompass more platforms and regions however, the public ECR storing the image started throttling. Despite efforts to mitigate this throttle by increasing the API limit and also distributing the image to multiple public ECRs, the throttling still occurs and we are currently unable to determine the precise reason why. 

After discussion, we have decided to revert from using image containers and explore other solutions to mitigate these transient issues such as caching. 

*Description of changes:*
Stop using image containers in the github runners and add back the retry logic to install Terraform and other dependencies.

Test run: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10171443800


.github/workflows/java-ec2-asg-e2e-test.yml
```diff
@@ -95,6 +93,13 @@ jobs:
             echo GET_CW_AGENT_RPM_COMMAND="wget -O cw-agent.rpm https://amazoncloudwatch-agent-${{ env.E2E_TEST_AWS_REGION }}.s3.${{ env.E2E_TEST_AWS_REGION }}.amazonaws.com/amazon_linux/amd64/latest/amazon-cloudwatch-agent.rpm" >> $GITHUB_ENV
           fi
 
+      - name: Set up terraform
```
majanjua-amzn, line 96: Run this and terraform init at the start, similar to gradle, then retry one or both as needed. If the time between the two attempts is small, add a 30s sleep to start

harrryr: The situation is different. For Gradle, we do this because we have a block of time while the Terraform deployment is being done (~1-2 minutes). For Terraform, we have to deploy Terraform Deploy almost immediately after the Terraform is setup, so it makes no sense to make two steps and add a 30s sleep. That's just retrying with extra steps.

harrryr: Example PR: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10170213176/job/28128848180

Check the time interval between runs

majanjua-amzn: Let's at least set the sleep time for this step to something higher than 10s since this is an installation that could fail due to connection issues.

majanjua-amzn (approved):

PR #162, opened 2024-08-07
Check if sample app is up

harrryr:
*Issue description:*
After removing the public endpoints, we removed the logic for checking if the sample application is up since we are unable to ping the sample app through the workflow. However, recently there have been increasing failures in EC2s due to the sample application not generating telemetry. This PR aims to resolve this issue by adding the logic back

*Description of changes:*
Check if the sample application is up after building it in Terraform. If the sample application is unavailable, then throw exit 1 to indicate that the Terraform Deployment was a failure. This will trigger the e2e workflow to re-instantiate another EC2 instance and deploy the sample application again. 

For ASGs, since the ASG creates the EC2 instance we need a different way to check if the sample application is available or not since throwing exit 1 will not cause a failure. Will push this PR first then apply implementation for ASG if we still noticeable improvement

*Rollback procedure:*
Revert PR

Test run:
- [Run](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10291998960) to check if EC2s will pass after new implementation: 
- [Run](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10291376482/job/28483485792)to check if Terraform will fail and retry if sample application is not available (Did this by changing the port to an invalid one)

Test run after updating: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10292445294/job/28486870614

Final Test run: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10309116932/job/28538036319

### DIFF
--- a/terraform/java/ec2/asg/main.tf
+++ b/terraform/java/ec2/asg/main.tf
@@ -229,6 +229,8 @@ resource "null_resource" "remote_service_setup" {
         sleep 10
       done
 
+      echo "Successfully connected to remote endpoint"
+
       EOF
     ]
   }
--- a/terraform/java/ec2/asg/main.tf
+++ b/terraform/java/ec2/asg/main.tf
@@ -219,7 +219,7 @@ resource "null_resource" "remote_service_setup" {
       # Check if the application is up. If it is not up, then exit 1.
       attempt_counter=0
       max_attempts=30
-      until $(curl --output /dev/null --silent --head --fail --max-time 300 $(echo "http://localhost:8080/healthcheck" | tr -d '"')); do
+      until $(curl --output /dev/null --silent --head --fail --max-time 5 $(echo "http://localhost:8080/healthcheck" | tr -d '"')); do
         if [ $attempt_counter -eq $max_attempts ];then
           echo "Failed to connect to endpoint."
           exit 1
--- a/terraform/java/ec2/asg/main.tf
+++ b/terraform/java/ec2/asg/main.tf
@@ -152,6 +152,7 @@ resource "aws_autoscaling_group" "asg" {
   max_size = 1
   launch_configuration = aws_launch_configuration.launch_configuration.name
   vpc_zone_identifier = [data.aws_subnets.default_subnets.ids.0]
+  health_check_type = "EC2"
 }
 
 resource "aws_instance" "remote_service_instance" {
@@ -214,6 +215,19 @@ resource "null_resource" "remote_service_setup" {
 
       # The application needs time to come up and reach a steady state, this should not take longer than 30 seconds
       sleep 30
+
+      # Check if the application is up. If it is not up, then exit 1.
+      attempt_counter=0
+      max_attempts=30
+      until $(curl --output /dev/null --silent --head --fail $(echo "http://localhost:8080/healthcheck" | tr -d '"')); do
+        if [ $attempt_counter -eq $max_attempts ];then
+          echo "Failed to connect to endpoint. "
+          exit 1
+        fi
+        printf '.'
+        attempt_counter=$(($attempt_counter+1))
+        sleep 10
+      done
 
       EOF
     ]
--- a/terraform/java/ec2/asg/main.tf
+++ b/terraform/java/ec2/asg/main.tf
@@ -219,7 +219,7 @@ resource "null_resource" "remote_service_setup" {
       # Check if the application is up. If it is not up, then exit 1.
       attempt_counter=0
       max_attempts=30
-      until $(curl --output /dev/null --silent --head --fail $(echo "http://localhost:8080/healthcheck" | tr -d '"')); do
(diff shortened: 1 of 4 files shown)


terraform/java/ec2/default/main.tf
```diff
@@ -138,6 +138,19 @@ resource "null_resource" "main_service_setup" {
...
+      max_attempts=30
+      until $(curl --output /dev/null --silent --head --fail $(echo "http://localhost:8080" | tr -d '"')); do
+        if [ $attempt_counter -eq $max_attempts ];then
+          echo "Failed to connect to endpoint. "
+          exit 1
+        fi
+        printf '.'
+        attempt_counter=$(($attempt_counter+1))
+        sleep 10
+      done
```
majanjua-amzn, line 152: We have an opportunity to make this logging more meaningful instead of just printing a line of `.`

terraform/java/ec2/asg/main.tf
```diff
@@ -152,6 +152,7 @@ resource "aws_autoscaling_group" "asg" {
   max_size = 1
   launch_configuration = aws_launch_configuration.launch_configuration.name
   vpc_zone_identifier = [data.aws_subnets.default_subnets.ids.0]
+  health_check_type = "EC2"
```
majanjua-amzn, line 155: What does this do?

harrryr: Tells the ASG to check if EC2 instances are not up, and if not then terminate and create new one

majanjua-amzn: Does that create a loop if there is a mistake in the code that would lead to the ec2 instance never coming up? Does the terraform behave normally in this case?

I would recommend trying to put a syntax error in the launch code that starts the instrumentation and sample app and seeing what happens

terraform/java/ec2/asg/main.tf
```diff
@@ -215,6 +216,19 @@ resource "null_resource" "remote_service_setup" {
...
 
+      # Check if the application is up. If it is not up, then exit 1.
+      attempt_counter=0
+      max_attempts=30
+      until $(curl --output /dev/null --silent --head --fail $(echo "http://localhost:8080/healthcheck" | tr -d '"')); do
+        if [ $attempt_counter -eq $max_attempts ];then
+          echo "Failed to connect to endpoint. "
+          exit 1
+        fi
+        printf '.'
```
majanjua-amzn, line 227: Didn't update here

terraform/java/ec2/default/main.tf
```diff
@@ -138,6 +138,19 @@ resource "null_resource" "main_service_setup" {
...
 
+      # Check if the application is up. If it is not up, then exit 1.
+      attempt_counter=0
+      max_attempts=30
+      until $(curl --output /dev/null --silent --head --fail $(echo "http://localhost:8080" | tr -d '"')); do
+        if [ $attempt_counter -eq $max_attempts ];then
+          echo "Failed to connect to endpoint. "
+          exit 1
+        fi
+        printf 'Attempting to connect to the endpoint. Tried $attempt_counter out of $max_attempts'
```
majanjua-amzn, line 149: use echo, `printf` will append and not start a new line I think

majanjua-amzn: And specify which endpoint (main vs remote)

terraform/java/ec2/asg/main.tf
harrryr: Yeah already tried that. Unfortunately code in the launch template doesn't cause the instance to fail to start running

terraform/java/ec2/default/main.tf
harrryr: Done

terraform/java/ec2/asg/main.tf
harrryr: Done

terraform/java/ec2/asg/main.tf
```diff
@@ -215,6 +216,19 @@ resource "null_resource" "remote_service_setup" {
       # The application needs time to come up and reach a steady state, this should not take longer than 30 seconds
       sleep 30
 
+      # Check if the application is up. If it is not up, then exit 1.
+      attempt_counter=0
+      max_attempts=30
+      until $(curl --output /dev/null --silent --head --fail $(echo "http://localhost:8080/healthcheck" | tr -d '"')); do
```
majanjua-amzn, line 222: Almost forgot, can you enforce a --max-time or something of the sort on this? We previously saw this retry take a total of 1h15 randomly instead of the expected 30*10. Put an appropriate limit using one or more of the time-related options for curl

harrryr: Done

majanjua-amzn (approved):

majanjua-amzn (approved):

PR #178, opened 2024-08-22
Add A Retry For Unstarted Jobs

harrryr:
*Issue description:*
Github jobs sometime fails to start like this job [here](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10404651182/job/28813573511). This is due to internal error with Github that we have no control over, so we are implementing a retry mechanism for such cases

*Description of changes:*
Currently we have a `-canary.yml` workflow which initializes enablement test run for each region using a matrix. We are adding a second layer between the `-canary.yml` and `e2e-test.yml` workflow called `-retry.yml` that checks if the `e2e-test.yml` workflow failed due to the above mentioned error, and if so, run a new test. 

- Removed references to `e2e`, replaced with `enablement` if needed
- Moved concurrency group to `-retry.yml`, updated concurrency group to have branch name and also manually inputted workflow name. This is because if the different enablement tests are defined in the same workflow, such as when we run tests or in release testing, then the concurrency group start blocking each other and make the overall run slower to finish
- Moved public metric job to `-retry.yml`

*Rollback procedure:*
Revert

*Test run:*
Successful EKS: [Run](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10517002486)
Successful EC2: [Run](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10517004283)
Successful K8s: [Run](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10517001679)

First enablement attempt failed: [Run](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10516956484)
Second enablement attempt failed too: [Run](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10516286398)
First retry attempt failed: [Run](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10516289584)
First enablement attempt and first retry attempt failed: [Run](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10516480860)
Both enablement attempts and first retry attempt failed: [Run](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10516487372)

- Checked that second jobs run when first jobs fail
- Checked that this logic `validation-result: ${{ needs.java-ec2-asg-attempt-1.outputs.validation-result || needs.java-ec2-asg-attempt-2.outputs.validation-result }}` outputs the success value when either the enablement attempt 1 or attempt 2 succeeds regardless of whether it was publish attempt 1 or 2


.github/workflows/java-ec2-asg-retry.yml
```diff
@@ -0,0 +1,57 @@
...
+      caller-workflow-name: ${{ inputs.caller-workflow-name }}
+
+  java-ec2-asg-attempt-2:
+    needs: [ java-ec2-asg-attempt-1 ]
+    if: ${{ needs.java-ec2-asg-attempt-1.outputs.job-started != 'true' }}
+    uses: ./.github/workflows/java-ec2-asg-test.yml
+    secrets: inherit
+    with:
+      aws-region: ${{ inputs.aws-region }}
+      caller-workflow-name: ${{ inputs.caller-workflow-name }}
```
majanjua-amzn, line 37: Each attempt will generate a different testing ID right?

.github/workflows/enablement-test-publish-result.yml
```diff
@@ -1,4 +1,4 @@
-name: Application Signals Enablement E2E Testing - Publish Result
+name: Application Signals Enablement Testing - Publish Result
```
majanjua-amzn, line 1: Something like this because we know we're in App Signals testing from the name of the repo, just makes it harder to read with a long name in the actions page
```suggestion
Publish Canary Test Result
```

.github/workflows/java-ec2-asg-retry.yml
```diff
@@ -0,0 +1,57 @@
+## Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+## SPDX-License-Identifier: Apache-2.0
+
+# This is a reusable workflow for running the Enablement test for App Signals.
+# It is meant to be called from another workflow.
+# Read more about reusable workflows: https://docs.github.com/en/actions/using-workflows/reusing-workflows#overview
+name: Java EC2 asg Retry
```
majanjua-amzn, line 7: nit:
```suggestion
name: Java EC2 ASG Retry
```

.github/workflows/java-ec2-asg-retry.yml
```diff
@@ -0,0 +1,57 @@
...
+
+  publish-metric-attempt-1:
+    needs: [ java-ec2-asg-attempt-1, java-ec2-asg-attempt-2 ]
+    if: always()
+    uses: ./.github/workflows/enablement-test-publish-result.yml
+    secrets: inherit
+    with:
+      aws-region: ${{ inputs.aws-region }}
+      caller-workflow-name: ${{ inputs.caller-workflow-name }}
+      validation-result: ${{ needs.java-ec2-asg-attempt-1.outputs.validation-result || needs.java-ec2-asg-attempt-2.outputs.validation-result }}
```
majanjua-amzn, line 47: Need some proof in PR description this works in all cases

.github/workflows/java-ec2-default-test.yml
```diff
@@ -239,7 +249,6 @@ jobs:
           aws-region: ${{ env.E2E_TEST_AWS_REGION }}
 
       - name: Save test results
-        if: always()
```
majanjua-amzn, line 242: Why? It wasn't removed in the ASG version of this test.

.github/workflows/java-eks-retry.yml
```diff
@@ -0,0 +1,66 @@
...
+        type: string
+      test-cluster-name:
+        required: true
+        type: string
+      caller-workflow-name:
+        required: true
+        type: string
+
+concurrency:
+  group: 'java-eks-${{ inputs.aws-region }}-${{ github.ref_name }}'
```
majanjua-amzn, line 22: Motivation behind replacing `'${{ github.workflow }} @ ${{ inputs.aws-region }}'`? What does this value look like now?

We could also do something like `'${{inputs.test-cluster-name}} @ ${{ inputs.aws-region}}'`, because cluster is the limiting factor here.

.github/workflows/enablement-test-publish-result.yml
majanjua-amzn, line 1: Can this be an action instead of a workflow? With all the variables used as inputs to the action instead so that it doesn't need access to secrets

.github/workflows/python-ec2-asg-test.yml
```diff
@@ -1,7 +1,7 @@
 ## Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 ## SPDX-License-Identifier: Apache-2.0
 
-# This is a reusable workflow for running the Python E2E Canary test for Application Signals.
+# This is a reusable workflow for running the Python Enablement Canary test for Application Signals.
 # It is meant to be called from another workflow.
 # Read more about reusable workflows: https://docs.github.com/en/actions/using-workflows/reusing-workflows#overview
 name: Python EC2 Asg Use Case
```
majanjua-amzn, line 7: nit: I know it's not part of your changes but while you're at it
```suggestion
name: Python EC2 ASG Use Case
```

majanjua-amzn (commented):
Looks great! Some questions and minor comments

Might be unconventional, but we could consider renaming any workflows that are not called directly to something with a prefix `z` so that they appear at the bottom of the list in the actions page.

.github/workflows/java-ec2-asg-retry.yml
harrryr: Yes, `${{ github.run_id }}` this should ensure that each have different test ID

.github/workflows/java-ec2-asg-retry.yml
harrryr: Ran test to verify this. added extra comment in PR description

.github/workflows/java-ec2-default-test.yml
harrryr: Remove by accident, readded

.github/workflows/java-eks-retry.yml
harrryr: Mentioned the motivation in the PR description. 

I could add your suggestion for EKS cases, but your suggestion is difficult to implement in K8s case so we will still need `${{ github.ref_name }}` and `java-k8s`

.github/workflows/enablement-test-publish-result.yml
harrryr: I'm not 100% sure, but I think it will probably work. 
Why do we want to make it into an action though? We intentionally use publish metric as a separate job, so writing it as a reusable workflow seems to make a lot of sense to me

majanjua-amzn: You're right, either way works but your implementation fits better with our current design.

.github/workflows/java-eks-retry.yml
majanjua-amzn: Works either way, understand the motivation now. Resolving comment

majanjua-amzn (approved):

majanjua-amzn (approved):

PR #180, opened 2024-08-23
Check if ASG instances are ready before proceeding to validation

harrryr:
*Issue description:*
This PR aims to address 2 issues: 
1. EC2 Instance created by ASG was unhealthy ([Link](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10410519220/job/28832546846))
2. Traffic Generator took too long to become ready ([Link](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10405692731/job/28817032152))

Both cases are happening because we are proceeding to log validation immediately after we deploy the asg without checking whether the EC2 instances launched by the asg is ready or not.

*Description of changes:*
1. Added a step to wait till the EC2 instance status is 'ok' before proceeding to validation step. For the first issue, it takes around 12 minutes for the ASG to refresh unhealthy instances so the wait time was set to 15 minutes. This issue only occurs in EC2 ASG cases so it is acceptable to sometimes have the EC2 case go over 15 minutes as it will not block any subsequent runs. 
2. Reduced health grace period time to 3 minutes. This is to reduce the 12 minute mentioned above to 10 minute.
4. Increased log validation time from ~2.5 minutes to ~3.3 minutes. Change 1 usually takes around 1.5 minutes if there are no issues, so giving 2.3 more minutes for the traffic generator to become available is enough time to resolve issue 2. (Although this isn't really neccessary because change 1 should resolve the issue anyways, added it as a fail-safe)

*Test Run*:
- Tested in private account to check whether reducing the grace period to 3 minutes will cause any issue.
- Test [run](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10529856233) for ASG

*Rollback procedure:*
Revert PR


.github/workflows/java-ec2-asg-e2e-test.yml
```diff
@@ -155,12 +155,33 @@ jobs:
...
+            if [ "$status_ok" == "true" ]; then
+              echo "The instance is ready."
+              echo "INSTANCE_ID=$instance_id" >> $GITHUB_ENV
+              break
+            fi
+          
+            if [ $retry_counter -eq $max_retry ]; then
+              echo "Instance did not become ready"
+              exit 1
+            fi
```
majanjua-amzn, line 176: Let's log the instance ID and current status

.github/workflows/java-ec2-asg-e2e-test.yml
```diff
@@ -155,12 +155,33 @@ jobs:
...
           done
 
+      - name: Wait for ASG instance to be ready
+        run: |
+          # Wait for up to 15 minutes
+          retry_counter=0
+          max_retry=90
+          while [ $retry_counter -lt $max_retry ]; do
+            instance_id=$(aws autoscaling describe-auto-scaling-groups --auto-scaling-group-names ec2-single-asg-${{ env.TESTING_ID }} --region ${{ env.E2E_TEST_AWS_REGION }} --query "AutoScalingGroups[].Instances[0].InstanceId" --output text)
+            status_ok=$(aws ec2 describe-instance-status --instance-ids $instance_id | jq '(.InstanceStatuses[] | select(.InstanceStatus.Status == "ok" and .SystemStatus.Status == "ok")) != {}')
```
majanjua-amzn, line 165: Can we get throttled here? Calling this EC2 and ASG API every 10 seconds seems reasonable, just curious. For safety we can do 45 x 20s instead of 90 x 10s if you find that there's a concern, for example.

.github/workflows/java-ec2-asg-e2e-test.yml
```diff
@@ -155,12 +155,33 @@ jobs:
...
+            
+            if [ "$status_ok" == "true" ]; then
+              echo "The instance is ready."
+              echo "INSTANCE_ID=$instance_id" >> $GITHUB_ENV
```
majanjua-amzn, line 169: Let's be clearer here for future engineers
```suggestion
              echo "MAIN_SERVICE_INSTANCE_ID=$instance_id" >> $GITHUB_ENV
```

validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
```diff
@@ -36,7 +36,7 @@
 
 @Log4j2
 public class CWLogValidator implements IValidator {
-  private static int DEFAULT_MAX_RETRY_COUNT = 15;
+  private static int DEFAULT_MAX_RETRY_COUNT = 20;
```
majanjua-amzn, line 39: Can we drop the metric validation retry count down from 80 to something more reasonable? Mostly unrelated to this change but worth keeping our limits in check

.github/workflows/java-ec2-asg-e2e-test.yml
harrryr: Checked briefly but not a concern: https://docs.aws.amazon.com/ec2/latest/devguide/ec2-api-throttling.html

validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
harrryr: Also reduced it to 30. Should be 6 minutes ish now

majanjua-amzn (approved):

PR #191, opened 2024-08-27
Patch K8s on EC2 OS Part 1

harrryr:
*Issue description:*
The EC2 AMI has an expiry date. Since our K8s on EC2 are designed to last indefinitely, eventually we will have the update the AMI. This PR aims to automate updating the AMIs without any manual interference. This will involve creating a new EC2 instance and taking down the old one

[Part 1](https://github.com/aws-observability/aws-application-signals-test-framework/pull/191): Create a new K8s on EC2 and test it
[Part 2](https://github.com/aws-observability/aws-application-signals-test-framework/compare/k8s-automation-part-2?expand=1): Verify that the instances were created properly and no impact on existing K8s, then replace the K8s

*Description of changes:*
**k8s-patch-os-matrix.yml**
Define all K8s on EC2s instances including the ones for canary as well as release testing. Any new K8s on EC2s will be defined here. It will update the AMI every month

**k8s-patch-os-jobs.yml**
1. Run a script that was copied and modified from [here](https://github.com/aws-observability/application-signals-demo/blob/main/scripts/k8s/appsignals/setup-k8s-demo.sh). Creates new key pair and two ec2 instances, which will have kubernetes installed
2. Store the newly created public ip and private ssh key in a secret key called `-pending-value`. Do not store it in the main secret yet because we want to make sure that our K8s test passes before updating the main secret key. 
3. Run K8s E2E test on newly created instance and verify it succeeds. All tests have been figured to use the release artifacts, including the ones for release testing since we don't want the test to fail due to invalid staging artifacts
4. Save main secret values onto a key called `-temporary-storage` then replace the main secret values with the newly created ones. This is so that in case of issues with the new k8s, we can find the original values and revert it back

Testing Id for K8s was updated to give a random value because all the K8s tests run under the same run id and job id is not retrievable in Github Action. Needed a way to distinguish between different runs or the validators will not be able to tell which telemetry to retrieve

Updated the ssm parameter key to use testing id instead of repository to make it unique to each run

*Rollback procedure:*
Revert PR, newly created EC2 instances will not affect canary. Make sure that the main secret keys were not altered. 

*Test*
Test run: https://github.com/harrryr/aws-application-signals-test-framework/actions/runs/10574639963
Test run on canary k8s: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10593567350

### DIFF
--- a/.github/workflows/java-k8s-test.yml
+++ b/.github/workflows/java-k8s-test.yml
@@ -14,6 +14,9 @@ on:
       caller-workflow-name:
         required: true
         type: string
+      caller-repository:
+        required: false
+        type: string
       adot-image-name:
         required: false
         type: string
@@ -33,6 +36,7 @@ permissions:
 env:
   E2E_TEST_AWS_REGION: ${{ inputs.aws-region }}
   CALLER_WORKFLOW_NAME: ${{ inputs.caller-workflow-name }}
+  CALLER_REPOSITORY: ${{ inputs.caller-repository }}
   ADOT_IMAGE_NAME: ${{ inputs.adot-image-name }}
   CW_AGENT_OPERATOR_TAG: ${{ inputs.cw-agent-operator-tag }}
   E2E_TEST_ACCOUNT_ID: ${{ secrets.APPLICATION_SIGNALS_E2E_TEST_ACCOUNT_ID }}
@@ -55,7 +59,7 @@ jobs:
         run: echo "job-started=true" >> $GITHUB_OUTPUT
 
       - name: Generate testing id
-        run: echo TESTING_ID="${{ env.E2E_TEST_AWS_REGION }}-${{ github.run_id }}-${{ github.run_number }}" >> $GITHUB_ENV
+        run: echo TESTING_ID="${{ github.run_id }}-${{ github.run_number }}-${RANDOM}" >> $GITHUB_ENV
 
       - uses: actions/checkout@v4
         with:
@@ -81,7 +85,7 @@ jobs:
           role-to-assume: arn:aws:iam::${{ env.E2E_TEST_ACCOUNT_ID }}:role/${{ env.E2E_TEST_ROLE_NAME }}
           aws-region: us-east-1
 
-      - name: Retrieve account
+      - name: Retrieve Secrets
         uses: aws-actions/aws-secretsmanager-get-secrets@v1
         with:
           secret-ids: |
@@ -89,8 +93,22 @@ jobs:
             JAVA_MAIN_SAMPLE_APP_IMAGE, e2e-test/java-main-sample-app-image
             JAVA_REMOTE_SAMPLE_APP_IMAGE, e2e-test/java-remote-sample-app-image
             RELEASE_TESTING_ECR_ACCOUNT, e2e-test/${{ github.event.repository.name }}/java-k8s-release-testing-account
+
+      - name: Retrieve K8s EC2 Secrets
+        if: ${{ env.CALLER_WORKFLOW_NAME != 'k8s-os-patching' }}
+        uses: aws-actions/aws-secretsmanager-get-secrets@v1
+        with:
+          secret-ids: |
             MAIN_SERVICE_ENDPOINT, e2e-test/${{ github.event.repository.name }}/java-k8s-master-node-endpoint
             MASTER_NODE_SSH_KEY, e2e-test/${{ github.event.repository.name }}/java-k8s-ssh-key
+
+      - name: Retrieve K8s EC2 OS Patching Secrets
+        if: ${{ env.CALLER_WORKFLOW_NAME == 'k8s-os-patching' }}
+        uses: aws-actions/aws-secretsmanager-get-secrets@v1
+        with:
+          secret-ids: |
+            MAIN_SERVICE_ENDPOINT, e2e-test/${{ env.CALLER_REPOSITORY }}/java-k8s-master-node-endpoint-pending-value
+            MASTER_NODE_SSH_KEY, e2e-test/${{ env.CALLER_REPOSITORY }}/java-k8s-ssh-key-pending-value
 
       - name: Prepare and upload sample app deployment files
         working-directory: terraform/java/k8s/deploy/resources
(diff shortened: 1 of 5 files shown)


.github/workflows/k8s-patch-os-jobs.yml
```diff
@@ -0,0 +1,173 @@
+# The EC2 AMI for K8s needs to be updated to prevent it from becoming outdated. This workflow will run in the beginning of every month to replace the existing
+# instances with new ones
+name: Update K8s EC2 Instance
```
majanjua-amzn, line 3: This and the other job have the same name? Be more descriptive but stay concise.

.github/workflows/k8s-patch-os-jobs.yml
```diff
@@ -0,0 +1,173 @@
...
+
+on:
+  workflow_call:
+    inputs:
+      repo:
```
majanjua-amzn, line 8: nit:
```suggestion
      repo_name:
```

.github/workflows/k8s-patch-os-jobs.yml
```diff
@@ -0,0 +1,173 @@
...
+            "e2e-test/${{ env.REPO }}/${{ env.LANGUAGE }}-k8s-master-node-endpoint-temporary-storage"
+            "e2e-test/${{ env.REPO }}/${{ env.LANGUAGE }}-k8s-ssh-key-temporary-storage"
+          )
+
+          for SECRET_NAME in "${SECRET_NAMES[@]}"; do
+            echo "Checking secret: $SECRET_NAME"
+
+
+            if aws secretsmanager describe-secret --secret-id $SECRET_NAME; then
+              echo "Success: The secret exists."
```
majanjua-amzn, line 68:
```suggestion
              echo "Success: Secret \"<INSERT THE SECRET NAME>\" already exists."
```

.github/workflows/k8s-patch-os-jobs.yml
```diff
@@ -0,0 +1,173 @@
...
+            else
+              echo "Secret does not exist. Creating new secret"
```
majanjua-amzn, line 70:
```suggestion
              echo "Secret \"<INSERT THE SECRET NAME>\" does not exist. Creating it now"
```

.github/workflows/k8s-patch-os-jobs.yml
```diff
@@ -0,0 +1,173 @@
...
+          SECRET_NAMES=(
+            "e2e-test/${{ env.REPO }}/${{ env.LANGUAGE }}-k8s-master-node-endpoint-pending-value"
+            "e2e-test/${{ env.REPO }}/${{ env.LANGUAGE }}-k8s-ssh-key-pending-value"
+            "e2e-test/${{ env.REPO }}/${{ env.LANGUAGE }}-k8s-master-node-endpoint"
+            "e2e-test/${{ env.REPO }}/${{ env.LANGUAGE }}-k8s-ssh-key"
+            "e2e-test/${{ env.REPO }}/${{ env.LANGUAGE }}-k8s-master-node-endpoint-temporary-storage"
+            "e2e-test/${{ env.REPO }}/${{ env.LANGUAGE }}-k8s-ssh-key-temporary-storage"
+          )
+
+          for SECRET_NAME in "${SECRET_NAMES[@]}"; do
```
majanjua-amzn, line 63: Looping on `aws secretsmanager` APIs, will we get throttled?

.github/workflows/k8s-patch-os-jobs.yml
```diff
@@ -0,0 +1,173 @@
...
+    uses: ./.github/workflows/python-k8s-e2e-test.yml
+    secrets: inherit
+    with:
+      aws-region: 'us-east-1'
+      caller-workflow-name: 'k8s-os-patching'
+      caller-repository: ${{ inputs.repo }}
+
+  update-secrets:
+    needs: [ run-java-k8s-test, run-python-k8s-test ]
+    if: ${{ always() && (needs.run-java-k8s-test.result == 'success' || needs.run-python-k8s-test.result == 'success') }}
```
majanjua-amzn, line 117: Shouldn't this be an AND operation? Or done separately for each language?
```suggestion
    if: ${{ always() && (needs.run-java-k8s-test.result == 'success' && needs.run-python-k8s-test.result == 'success') }}
```

.github/workflows/k8s-patch-os-jobs.yml
```diff
@@ -0,0 +1,173 @@
...
+
+  run-python-k8s-test:
+    if: ${{ inputs.LANGUAGE == 'python' }}
+    needs: create-k8s-on-ec2
+    uses: ./.github/workflows/python-k8s-e2e-test.yml
+    secrets: inherit
+    with:
+      aws-region: 'us-east-1'
+      caller-workflow-name: 'k8s-os-patching'
+      caller-repository: ${{ inputs.repo }}
```
majanjua-amzn, line 113: How do we scale this for more languages? Just write more of these steps?

.github/workflows/k8s-patch-os-jobs.yml
```diff
@@ -0,0 +1,173 @@
...
+          
+          endpoint_secret=$(aws secretsmanager get-secret-value --secret-id "${CURRENT_SECRET_KEYS[0]}" --query SecretString --output text)
+          ssh_secret=$(aws secretsmanager get-secret-value --secret-id "${CURRENT_SECRET_KEYS[1]}" --query SecretString --output text)
+          
+          aws secretsmanager update-secret --secret-id "${STORAGE_SECRET_KEYS[0]}" --secret-string $endpoint_secret
+          aws secretsmanager update-secret --secret-id "${STORAGE_SECRET_KEYS[1]}" --secret-string "$ssh_secret"
+
+#      - name: Replace current endpoint and ssh key with the newly created ones
+#        run: |
+#          PENDING_SECRET_KEYS=(
```
majanjua-amzn, line 146: Why is this commented? Looks like leftover code

.github/workflows/python-k8s-test.yml
```diff
@@ -257,3 +275,10 @@ jobs:
         run: |
           terraform destroy -auto-approve \
             -var="test_id=${{ env.TESTING_ID }}"
+
+      - name: Delete deployment files
+        if: always()
+        continue-on-error: true
+        run: |
+          aws s3api delete-object --bucket aws-appsignals-sample-app-prod-${{ env.E2E_TEST_AWS_REGION }} --key python-frontend-service-depl-${{ env.TESTING_ID }}.yaml
+          aws s3api delete-object --bucket aws-appsignals-sample-app-prod-${{ env.E2E_TEST_AWS_REGION }} --key python-remote-service-depl-${{ env.TESTING_ID }}.yaml
```
majanjua-amzn, line 284: Going to need to add this to the cleanup script as well in case it's missed or it fails randomly

.github/workflows/util/setup-k8s.sh
majanjua-amzn, line 1: Do the instances created by this script have the SSM agent or whatever is needed to emit OS patching status installed? Can you point out which lines cover that? Ty

.github/workflows/java-k8s-test.yml
```diff
@@ -55,7 +61,7 @@ jobs:
         run: echo "job-started=true" >> $GITHUB_OUTPUT
 
       - name: Generate testing id
-        run: echo TESTING_ID="${{ env.E2E_TEST_AWS_REGION }}-${{ github.run_id }}-${{ github.run_number }}" >> $GITHUB_ENV
+        run: echo TESTING_ID="${{ github.run_id }}-${{ github.run_number }}-${RANDOM}" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
harrryr: It's built in, calling it generates a random number. Reason mentioned in the PR description

.github/workflows/k8s-patch-os-jobs.yml
harrryr: Hmm, the secret name is already called in line 64 `echo "Checking secret: $SECRET_NAME"`, but I guess it's more clear if we add it in the result message too

.github/workflows/k8s-patch-os-jobs.yml
harrryr: No, limit is 10,000 requests per second 
https://docs.aws.amazon.com/secretsmanager/latest/userguide/reference_limits.html

.github/workflows/k8s-patch-os-jobs.yml
harrryr: It needs to be an `||` operation because only k8s-java or k8s-python test will run depending on which language the instance is for.

.github/workflows/k8s-patch-os-jobs.yml
harrryr: Yes. I think this is the best we can do since reusuable workflows have to be hardcoded

.github/workflows/k8s-patch-os-jobs.yml
harrryr: Commented it out so that we can first check if the changes were as expected, then we can merged part 2 to uncomment this section of the code

.github/workflows/python-k8s-test.yml
harrryr: Ok

.github/workflows/util/setup-k8s.sh
harrryr: https://github.com/aws-observability/aws-application-signals-test-framework/pull/191/files#diff-3350f7ca55f76be841c1358d5761a465dcf4aacea8b4b74e143f978c9c6a51c5R90

.github/workflows/k8s-patch-os-jobs.yml
majanjua-amzn: Understood, ty

majanjua-amzn (approved):

PR #198, opened 2024-08-29
K8s OS Patch Cleanup 

harrryr:
*Issue description:*
We need a mechanism to cleanup old K8s on EC2 instances and also make sure that instances are terminated if the os patching process fails

*Description of changes:*
- Run a script everyday to clean up old K8s on EC2s
- Run a script to clean up k8s if the os patching workflow fails
- Emit metric for deprecation script

Checked the script in the s3 bucket to make sure that the appropriate instances are listed.

There are 76 K8s instances that needs to be terminated

Test run: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10686799902/job/29622912272

*Rollback procedure:*
Revert PR


.github/workflows/resource-cleanup.yml
```diff
@@ -53,3 +53,76 @@ jobs:
...
+      matrix:
+        instance: [
+          { repo_name: 'amazon-cloudwatch-agent-operator', ec2_name: 'cw-agent-operator-release', language: 'java' },
+          { repo_name: 'amazon-cloudwatch-agent-operator', ec2_name: 'cw-agent-operator-release', language: 'python' },
+          { repo_name: 'amazon-cloudwatch-agent', ec2_name: 'cw-agent-release', language: 'java' },
+          { repo_name: 'amazon-cloudwatch-agent', ec2_name: 'cw-agent-release', language: 'python' },
+          { repo_name: 'aws-otel-python-instrumentation', ec2_name: 'adot-python-release', language: 'python' },
+          { repo_name: 'aws-otel-java-instrumentation', ec2_name: 'adot-java-release', language: 'java' },
+          { repo_name: 'aws-application-signals-test-framework', ec2_name: 'python-canary', language: 'python' },
+          { repo_name: 'aws-application-signals-test-framework', ec2_name: 'java-canary', language: 'java' } ]
```
majanjua-amzn, line 69: This is duplicated, is there a way we can have one place for all these so we don't forget them?

.github/workflows/resource-cleanup.yml
```diff
@@ -53,3 +53,76 @@ jobs:
...
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Check if today is the first day of the month
+        id: check_day
+        run: |
+          echo "is_first_day=$(( $(date +%d) == 01 ))" >> $GITHUB_OUTPUT
```
majanjua-amzn, line 77: We only run all this logic once a month? Would seem safer to do a full check/cleanup every day and detect unneeded clusters

Does this necessarily run after the replace jobs? What happens if this fails or if the replacement jobs fail and need to be rerun? I think we should have this always run

.github/workflows/resource-cleanup.yml
```diff
@@ -53,3 +53,76 @@ jobs:
...
+        run: |
+          STORAGE_SECRET_KEYS=(
+            "e2e-test/${{ matrix.instance.repo_name }}/${{ matrix.instance.language }}-k8s-master-node-endpoint-temporary-storage"
+          )
+          endpoint_secret=$(aws secretsmanager get-secret-value --secret-id "${STORAGE_SECRET_KEYS[0]}" --query SecretString --output text)
+          
+          instance_name=$(aws ec2 describe-instances \
+            --filters "Name=ip-address,Values=$endpoint_secret" \
+            --query "Reservations[*].Instances[*].{Name:Tags[?Key=='Name']|[0].Value}" \
+            --output text)
```
majanjua-amzn, line 97: So we check the temp secrets for old clusters, then delete those?

We should be checking all instances associated with k8s clusters and cross referencing them with the clusters we keep for releases and canaries. If they're not in that list, then delete them.

majanjua-amzn (approved):

PR #208, opened 2024-09-04
[Node/EC2] Canary/release test creation

majanjua-amzn:
### Issue description
Introduce first iteration of EC2 canary and EC2/ASG release tests. Release tests are not yet created, as the first release needs to be completed for the mechanisms to be ready for the test to test them effectively.

Notes:
- ASG will not be run as a canary, but was included in the PR for review as a release test and both it and the EC2 default case will be finalized later as we continue the release of JS support. As such, it was included in the canary/cronjob workflow file during testing and removed for the purpose of the PR, hence the mismatch in commit IDs between the test runs and the PR.
- The canary only runs in IAD, as we have not currently released the instrumentation into a publicly accessible/regionally accessible distribution. It is only accessible by our IAD account at this point. Again, we will come back and update this as part of our canary/release testing finalization.

### Description of changes
- New EC2/ASG test covering log, metric, and trace validation
- Modified NodeJS sample app to properly generate InternalOperation code
- Update trace validator to provide extra, useful information regarding which key failed to match when there is a mismatch

### Rollback procedure
Revert PR

### Testing
- Test run showing both EC2 default and EC2 ASG pass: [link](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10710109928)
- Test run 2: [link](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10710723264/job/29698040394)

### DIFF
--- a/sample-apps/node/frontend-service/index.js
+++ b/sample-apps/node/frontend-service/index.js
@@ -77,19 +77,30 @@ app.get('/remote-service', (req, res) => {
   request.end();
 });
 
-app.get('/client-call', (req, res) => {
-  // Immediately respond to the client without waiting for the HTTP request to complete
-  res.status(202).send('/client-call called successfully');
-  console.log('/client-call called successfully')
+// The following logic serves as the async call made by the /client-call API
+let makeAsyncCall = false;
+setInterval(() => {
+  if (makeAsyncCall) {
+    makeAsyncCall = false;
+    console.log('Async call triggered by /client-call API');
 
-  const request = http.get('http://local-root-client-call', (rs) => {
-    rs.setEncoding('utf8');
-    rs.on('data', (result) => {
-      res.send(`/client-call response: ${result}`);
+    const request = http.get('http://local-root-client-call', (rs) => {
+      rs.setEncoding('utf8');
+      rs.on('data', (result) => {
+        res.send(`GET local-root-client-call response: ${result}`);
+      });
     });
-  });
-  request.on('error', (err) => {});
-  request.end();
+    request.on('error', (err) => {}); // Expected
+    request.end();
+  }
+}, 5000); // Check every 5 seconds
+
+app.get('/client-call', (req, res) => {
+  res.send('/client-call called successfully');
+  console.log('/client-call called successfully');
+
+  // Trigger async call to generate telemetry for InternalOperation use case
+  makeAsyncCall = true;
 });
 
 app.get('/mysql', (req, res) => {
--- a/terraform/node/ec2/asg/main.tf
+++ b/terraform/node/ec2/asg/main.tf
@@ -1,0 +1,281 @@
+# ------------------------------------------------------------------------
+# Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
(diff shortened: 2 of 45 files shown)


sample-apps/node/frontend-service/index.js
```diff
@@ -77,19 +77,30 @@ app.get('/remote-service', (req, res) => {
...
-  const request = http.get('http://local-root-client-call', (rs) => {
-    rs.setEncoding('utf8');
-    rs.on('data', (result) => {
-      res.send(`/client-call response: ${result}`);
-    });
-  });
-  request.on('error', (err) => {});
-  request.end();
+  // Trigger async call to generate telemetry for InternalOperation use case
+  makeAsyncCall = true;
```
pxaws, line 103: If you just want to trigger the InternalOperation, you don't need this. The variable couples the the sync and async calls together.

sample-apps/node/frontend-service/index.js
```diff
@@ -77,19 +77,30 @@ app.get('/remote-service', (req, res) => {
   request.end();
 });
 
+// The following logic serves as the async call made by the /client-call API
+let makeAsyncCall = false;
+setInterval(() => {
+  if (makeAsyncCall) {
```
pxaws, line 83: maybe no need to check makeAsyncCall. Can we just send the request every 3 seconds at regular interval?

majanjua-amzn: That's a good idea, but this is in line with the Java/Python implementation so I think this should be okay for consistency.

Let me know what you think

sample-apps/node/frontend-service/index.js
majanjua-amzn: It's also to ensure the trace segments for the initial call and the async call are connected.

terraform/node/ec2/asg/main.tf
```diff
@@ -0,0 +1,281 @@
...
+
+    # Export environment variables for instrumentation
+    tmux send-keys -t frontend 'export OTEL_METRICS_EXPORTER=none' C-m
+    tmux send-keys -t frontend 'export OTEL_TRACES_EXPORTER=otlp' C-m
+    tmux send-keys -t frontend 'export OTEL_AWS_APPLICATION_SIGNALS_ENABLED=true' C-m
+    tmux send-keys -t frontend 'export OTEL_AWS_APPLICATION_SIGNALS_EXPORTER_ENDPOINT=http://localhost:4316/v1/metrics' C-m
+    tmux send-keys -t frontend 'export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://localhost:4316/v1/traces' C-m
+    tmux send-keys -t frontend 'export OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=http/protobuf' C-m
+    tmux send-keys -t frontend 'export OTEL_EXPORTER_OTLP_METRICS_PROTOCOL=http/protobuf' C-m
+    tmux send-keys -t frontend 'export OTEL_NODE_DISABLED_INSTRUMENTATIONS=fs,dns,express' C-m
```
jj22ee, line 136: Can we leave comment here (and the other 3 places this is set) to explain why we add `express` to disabled instrumentation list to reduce the overhead in the testing? I recall there is some random order introduced by express that affects the validator.

majanjua-amzn: Added a comment above each of the tmux sections where this is used.

sample-apps/node/frontend-service/index.js
pxaws: Interesting. How do we know if an async operation is triggered by a sync operation? Especially when the async operation is triggered by `setInterval` which just execute periodically.

pxaws (approved):

sample-apps/node/frontend-service/index.js
majanjua-amzn: You're right, they are not connected in the telemetry. Regardless, I think we should follow the implementation of Java/Python for consistency

PR #190, opened 2024-08-27
Adding remote resources upgrading mysql

BlueLort:
Original PR: https://github.com/aws-observability/aws-application-signals-test-framework/pull/103

Copying description here.

Issue description:

When we assert `RemoteResourceType` and `RemoteResourceIdentifier` attributes in [EMF logs](https://github.com/aws-observability/aws-application-signals-test-framework/blob/main/validator/src/main/resources/expected-data-template/java/eks/aws-sdk-call-log.mustache#L30-L33), we are using Logs filter: `RemoteService` and `RemoteOperation` filter to fetch logs. However, it's not guaranteed that EMF log with `RemoteService` and `RemoteOperation` attributes will have `RemoteResourceType` and `RemoteResourceIdentifier`. The logs fetched with this generic filter might get more logs than expected and only asserting the first log event in the result.

Description of changes:

This change enhances the log filter by appending  `&& ($.RemoteResourceType = %EXAMPLE%) && ($.RemoteResourceIdentifier = %EXAMPLE%)` to the filter if they present in the template to get more specific log events.

Log Filter doc: https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/FilterAndPatternSyntax.html#regex-expressions

Log Filter used before this change (From E2E testing log):
```
2024-07-02T13:41:46.9400277Z 13:41:46.840 [main] INFO  com.amazon.aoc.validators.CWLogValidator - Filter Pattern for Log Search: { ($.Service = sample-application-eu-west-1-9762031914-112) && ($.Operation = "GET /aws-sdk-call") && ($.RemoteService = "AWS::S3") && ($.RemoteOperation = "GetBucketLocation") }
```
Log Filter used after this change (From E2E testing log):
```
2024-07-02T13:49:00.6162597Z 13:49:00.510 [main] INFO  com.amazon.aoc.validators.CWLogValidator - Filter Pattern for Log Search: { ($.Service = sample-application-eu-west-1-9762033464-113) && ($.Operation = "GET /aws-sdk-call") && ($.RemoteService = "AWS::S3") && ($.RemoteOperation = "GetBucketLocation") && ($.RemoteResourceType = %^AWS::S3::Bucket$%) && ($.RemoteResourceIdentifier = %^e2e-test-bucket-name-eu-west-1-9762033464-113$%) }
```

**Pending E2E testing run**

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #209, opened 2024-09-05
Enhance Logs E2E test and bump up MySQL library to 8.4

ektabj:
*Issue description:*

In this PR:
-  [Supports remoteResourceType and remoteResourceIdentifier filter when fetching from Logs](https://github.com/aws-observability/aws-application-signals-test-framework/pull/190/commits/4222c15566ebeba307c641cb9a8946bc46e00d1d)
- [migrating MySQL to 8.4](https://github.com/aws-observability/aws-application-signals-test-framework/pull/190/commits/bd6a8266a9a97ba0708873331a48fbe22fa0e1f7)
- Removing regex not needed in github E2E template

E2E test run: https://github.com/ektabj/aws-application-signals-test-framework/actions/runs/10720299089

Copying description from other PRs here.

Issue description:

When we assert `RemoteResourceType` and `RemoteResourceIdentifier` attributes in [EMF logs](https://github.com/aws-observability/aws-application-signals-test-framework/blob/main/validator/src/main/resources/expected-data-template/java/eks/aws-sdk-call-log.mustache#L30-L33), we are using Logs filter: `RemoteService` and `RemoteOperation` filter to fetch logs. However, it's not guaranteed that EMF log with `RemoteService` and `RemoteOperation` attributes will have `RemoteResourceType` and `RemoteResourceIdentifier`. The logs fetched with this generic filter might get more logs than expected and only asserting the first log event in the result.

Description of changes:

This change enhances the log filter by appending  `&& ($.RemoteResourceType = %EXAMPLE%) && ($.RemoteResourceIdentifier = %EXAMPLE%)` to the filter if they present in the template to get more specific log events.

Log Filter doc: https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/FilterAndPatternSyntax.html#regex-expressions

Log Filter used before this change (From E2E testing log):
```
2024-07-02T13:41:46.9400277Z 13:41:46.840 [main] INFO  com.amazon.aoc.validators.CWLogValidator - Filter Pattern for Log Search: { ($.Service = sample-application-eu-west-1-9762031914-112) && ($.Operation = "GET /aws-sdk-call") && ($.RemoteService = "AWS::S3") && ($.RemoteOperation = "GetBucketLocation") }
```
Log Filter used after this change (From E2E testing log):
```
2024-07-02T13:49:00.6162597Z 13:49:00.510 [main] INFO  com.amazon.aoc.validators.CWLogValidator - Filter Pattern for Log Search: { ($.Service = sample-application-eu-west-1-9762033464-113) && ($.Operation = "GET /aws-sdk-call") && ($.RemoteService = "AWS::S3") && ($.RemoteOperation = "GetBucketLocation") && ($.RemoteResourceType = %^AWS::S3::Bucket$%) && ($.RemoteResourceIdentifier = %^e2e-test-bucket-name-eu-west-1-9762033464-113$%) }
```


validator/src/main/resources/expected-data-template/java/eks/aws-sdk-call-log.mustache
```diff
@@ -29,8 +29,8 @@
   "Version": "^1$",
   "RemoteService": "AWS::S3",
   "RemoteOperation": "GetBucketLocation",
-  "RemoteResourceIdentifier": "^e2e-test-bucket-name-{{testingId}}$",
-  "RemoteResourceType": "^AWS::S3::Bucket$",
+  "RemoteResourceIdentifier": "e2e-test-bucket-name-{{testingId}}",
+  "RemoteResourceType": "AWS::S3::Bucket",
   "Telemetry.Source": "^ClientSpan$",
```
jerry-shao, line 34: I just realized that we don't need to update these templates because this regex pattern is not for searching logs using logs filter, but actually look up the log entires after getting them from logs.

The change in validator should be good enough, but it's ok to do exactly match here. I don't mind.

majanjua-amzn: I think you would need to change the syntax if you wanted to use regex instead of exact match. I think the current implementation is fine.

PR #217, opened 2024-09-06
[K8s] Remove testing for new language clusters

majanjua-amzn:
## Issue/Description of changes
New clusters were using the java test, but the java test is hardcoded to use the temporary java clusters and does not automatically detect that it needs to use the clusters meant for other languages (which makes sense). As such:
- Removed the testing for now since the tests can be run after the fact anyways
- Also added run attempt to the testing ID so that the jobs in a workflow can be run again without overlapping resource names

## Rollback procedure
Revert and rerun k8s os patching


XinRanZhAWS (approved):
LGTM

harrryr (approved):

PR #222, opened 2024-09-09
Enable ECS end to end test for Java and Python

zzhlogin:
*Issue description:*

*Description of changes:*
This PR enabled ECS end to end test for Java and Python including workflows, terraform and validation setups. The change follows same procedure for other platform: EC2, EKS and K8S.

*Rollback procedure:*
Revert this commit. No further revert action needed.

Example of successful workflows running:
Java: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10906916872
Python:  https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/10906916891/job/30269312152


validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
```diff
@@ -140,8 +140,7 @@ private Map<String, Object> getTrace() throws Exception {
       traceFilter += " AND (annotation.aws_local_service = \"local-root-client-call\" OR annotation.aws_local_service = \"local-root-client-call:80\")" ;
     } else {
       traceFilter += (String.format(" AND annotation.aws_local_operation = \"%s %s\"",
-              validationConfig.getHttpMethod().toUpperCase(),
-              validationConfig.getHttpPath()));
+              validationConfig.getHttpMethod().toUpperCase(), validationConfig.getHttpPath()));
```
bjrara, line 143: Doesn't seem to be a necessary changes in this PR.

validator/src/main/java/com/amazon/aoc/services/XRayService.java
```diff
@@ -26,7 +26,9 @@
 import java.util.Date;
 import java.util.List;
 import org.joda.time.DateTime;
+import lombok.extern.log4j.Log4j2;
 
+@Log4j2
```
bjrara, line 31: Why do we need `Log4j2` annotation?

terraform/python/ecs/variables.tf
```diff
@@ -0,0 +1,49 @@
...
+variable "ecs_cluster_name" {
+  default = "e2e-test-java"
+}
+
+variable "sample_app_name" {
+  default = ""
+}
+
+variable "sample_app_image" {
+  // public.ecr.aws/m0o7p6d1/aws-apm-java-springboot-main-service:latest
```
bjrara, line 34: Please help remove the comments for test images. Same to the other variable declarations.

terraform/java/ecs/resources/main-service.json.tpl
```diff
@@ -0,0 +1,125 @@
...
+        "awslogs-group": "/ecs/ecs-cwagent",
+        "awslogs-create-group": "true",
+        "awslogs-region": "${aws_region}",
+        "awslogs-stream-prefix": "ecs"
+      }
+    }
+  },
+  {
+    "name": "traffic-gen",
+    "image": "curlimages/curl:8.8.0",
```
bjrara, line 116: I'm wondering if using a public image is okay. I'll leave this question to the maintainers.

terraform/java/ecs/output.tf
```diff
@@ -0,0 +1,18 @@
...
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+# -------------------------------------------------------------------------
+
+output "sample_app_main_service_task_definition" {
+  value = aws_ecs_task_definition.main_service.arn
```
bjrara, line 17: I created this output because I thought it would be useful during trouble shooting, but wondering if it is the case.

If not, what should be put in the output?

terraform/java/ecs/variables.tf
```diff
@@ -0,0 +1,49 @@
...
+variable "ecs_cluster_name" {
+  default = "e2e-test-java"
+}
+
+variable "sample_app_name" {
+  default = ""
+}
+
+variable "sample_app_image" {
+  // public.ecr.aws/m0o7p6d1/aws-apm-java-springboot-main-service:latest
```
bjrara, line 34: Please help remove the comments for test images. Same to the other variable declarations.

.github/workflows/java-ecs-canary.yml
```diff
@@ -0,0 +1,27 @@
...
+  ecs:
+    strategy:
+      fail-fast: false
+      matrix:
+        aws-region: ['us-east-1']
+    uses: ./.github/workflows/java-ecs-retry.yml
+    secrets: inherit
+    with:
+      aws-region: ${{ matrix.aws-region }}
+      caller-workflow-name: 'appsignals-e2e-java-ecs-canary-test'
```
majanjua-amzn, line 27: If you intend on testing in IAD alone, then you don't need a matrix.
```suggestion
  ecs:
    uses: ./.github/workflows/java-ecs-retry.yml
    secrets: inherit
    with:
      aws-region: 'us-east-1'
      caller-workflow-name: 'appsignals-e2e-java-ecs-canary-test'
```

.github/workflows/java-ecs-test.yml
```diff
@@ -0,0 +1,280 @@
...
+
+      - name: Set Get ADOT Wheel command environment variable
+        run: |
+          if [ "${{ github.event.repository.name }}" = "aws-otel-java-instrumentation" ]; then
+            # Reusing the adot-main-build-staging-jar bucket to store the java wheel file
+            echo ADOT_INSTRUMENTATION_IMAGE_URI="${{ env.ADOT_IMAGE_NAME }}" >> $GITHUB_ENV
+          else
+            ADOT_INSTRUMENTATION_IMAGE_TAG=$(curl -s -I -L 'https://github.com/aws-observability/aws-otel-java-instrumentation/releases/latest' | grep -i Location | awk -F'/tag/' '{print $2}' | tr -d '\r')
+            echo ADOT_INSTRUMENTATION_IMAGE_URI="public.ecr.aws/aws-observability/adot-autoinstrumentation-java:$ADOT_INSTRUMENTATION_IMAGE_TAG" >> $GITHUB_ENV
+          fi
```
majanjua-amzn, line 125: I assume it's not a WHL/wheel file? Name of the step and comments seem outdated, leading to confusion

validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
zzhlogin: Changed back to original format.

validator/src/main/java/com/amazon/aoc/services/XRayService.java
zzhlogin: It was used for debugging purpose. Removed.

terraform/python/ecs/variables.tf
zzhlogin: Done

terraform/java/ecs/resources/main-service.json.tpl
zzhlogin: Change to use "wget" instead without image needed to avoid security concern.

terraform/java/ecs/output.tf
zzhlogin: Similar as EKS test, I don't think we need a output, removed the `output.tf` file.

terraform/java/ecs/variables.tf
zzhlogin: Done

.github/workflows/java-ecs-canary.yml
zzhlogin: For now, we only want to test in IAD, but I think it will be better to use matrix, so that it will be convenient to add more regions if we want to test in the future.

.github/workflows/java-ecs-test.yml
zzhlogin: Thanks for the check. Updated to the right title.

bjrara (approved):
Overall looks good to me. Do we want to hold the PR until the artifacts are released?

validator/src/main/java/com/amazon/aoc/services/XRayService.java
```diff
@@ -26,6 +26,7 @@
 import java.util.Date;
 import java.util.List;
 import org.joda.time.DateTime;
+import lombok.extern.log4j.Log4j2;
```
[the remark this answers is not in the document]
majanjua-amzn: +1, looks like leftover change from testing

validator/src/main/resources/validations/java/ecs/log-validation.yml
```diff
@@ -0,0 +1,4 @@
+-
+  validationType: "cw-log"
+  callingType: "none"
+  expectedLogStructureTemplate: "JAVA_ECS_HC_CALL_LOG"
```
majanjua-amzn, line 4: If we're only doing one API call, then we can merge the validation files into one. Definitely not needed, having them separate is also valid so that they can be easier to debug, but just letting you know it's an option:
```suggestion
-
  validationType: "cw-log"
  callingType: "none"
  expectedLogStructureTemplate: "JAVA_ECS_HC_CALL_LOG"
-
  validationType: "cw-metric"
  callingType: "none"
  expectedMetricTemplate: "JAVA_ECS_HC_CALL_METRIC"
-
  validationType: "trace"
  callingType: "none"
  expectedTraceTemplate: "JAVA_ECS_HC_CALL_TRACE"
```

validator/src/main/resources/validations/java/ecs/trace-validation.yml
```diff
@@ -0,0 +1,6 @@
+-
+  validationType: "trace"
+  callingType: "none"
+  httpMethod: "get"
+  httpPath: "/"
```
majanjua-amzn, line 5: Why do we need httpMethod/httpPath if we have `none` callingType?

validator/src/main/resources/expected-data-template/python/ecs/hc-trace.mustache
```diff
@@ -0,0 +1,33 @@
...
+  {
+    "name": "^{{serviceName}}$",
+    "http": {
+      "request": {
+        "url": "^http://localhost:8080/$",
+        "method": "^GET$"
+      },
+      "response": {
+        "status": "^200$"
+      }
```
majanjua-amzn, line 11: Let's not validate response status, there are many reasons this can come out different and we don't generally care that the response status is different so long as the rest of the telemetry is fine.

validator/src/main/resources/expected-data-template/python/ecs/hc-metric.mustache
```diff
@@ -0,0 +1,74 @@
...
+      value: GET /
+    -
+      name: Service
+      value: {{serviceName}}
+    -
+      name: Environment
+      value: ecs:{{platformInfo}}
+
+-
+  metricName: Latency
```
majanjua-amzn, line 16: We only have 2 metric aggregations for ECS?

terraform/python/ecs/variables.tf
```diff
@@ -0,0 +1,46 @@
...
+# permissions and limitations under the License.
+# -------------------------------------------------------------------------
+
+
+variable "test_id" {
+  default = "dummy-123"
+}
+
+variable "aws_region" {
+  default = "us-west-2"
```
majanjua-amzn, line 22: Enforce the input by allowing it to fail if it's not provided
```suggestion
  default = "<e.g. us-east-1>"
```

validator/src/main/resources/validations/java/ecs/trace-validation.yml
zzhlogin: I found the httpMethod and httpPath is still being used [here](https://github.com/aws-observability/aws-application-signals-test-framework/blob/442f8fbab821d53222bc87e5e4e23c5e71d72d93/validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java#L143) during trace validation even callingType is none.

validator/src/main/resources/validations/java/ecs/log-validation.yml
zzhlogin: Got it. Thanks for the info. Let's continue to separate them follow the other platforms.

validator/src/main/resources/expected-data-template/python/ecs/hc-trace.mustache
zzhlogin: Updated

terraform/python/ecs/variables.tf
zzhlogin: Updated

validator/src/main/java/com/amazon/aoc/services/XRayService.java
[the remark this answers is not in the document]
zzhlogin: Removed

majanjua-amzn (approved):

PR #249, opened 2024-09-24
[Python/EKS] Fix credential interpolation issue in Enablement Canary Test

jerry-shao:
*Issue description:*

It's a known issue that RDS managed credential could generate a credential with special symbols like: `XXX$$XXX`. When we pass it to the Sample App via terraform template, it could cause string interpolation in bash and the credential got translated to `XXX<PID>XXX` which will be incorrect.

*Description of changes:*

This change convert RDS credentials to base64 first and then decode it in the Python Sample App. Re-run Python Sample App Image build workflow is required after merging this PR.

We have verified this solution in Java: https://github.com/aws-observability/aws-application-signals-test-framework/pull/244

*Rollback procedure:*

Yes.

We need to re-run Python Sample App Image build workflow as well after revert.

Example execution: https://github.com/jerry-shao/aws-application-signals-test-framework/actions/runs/11014255259/job/30584477033


sample-apps/python/django_frontend_service/frontend_service_app/views.py
```diff
@@ -2,6 +2,7 @@
 ## SPDX-License-Identifier: Apache-2.0
 import logging
 import os
+import base64
```
majanjua-amzn, line 5: Do we need to add this to `requirements.txt` and `ec2-requirements.txt`?

jerry-shao: @majanjua-amzn It's a built-in library from Python: https://docs.python.org/3/library/base64.html

The test run also confirmed it can import this library without add it as a dependency.

majanjua-amzn (approved):

PR #274, opened 2024-10-04
[K8s/Patching] Add jitter to K8s-on-EC2 cluster creation

majanjua-amzn:
### Issue
K8s OS patching workflow does not succeed some jobs because of throttling when trying to create EC2 instances

### Description of changes
- Added jitter_index to input
- Added sleep based on jitter index
- Each job will now sleep 20 seconds for each job before it in the list, ensuring they run somewhat sequentially and not all at the same time

Note: The github scheduler/worker assignment may still lead to some of the jobs doing things at the same time, but the risk of this causing throttling should be low

### Testing
https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/11172296461

### Rollback procedure
Revert PR and rerun the workflow

### DIFF
--- a/.github/workflows/k8s-patch-os-jobs.yml
+++ b/.github/workflows/k8s-patch-os-jobs.yml
@@ -5,6 +5,9 @@ name: K8s Patch OS Job
 on:
   workflow_call:
     inputs:
+      jitter_index:
+        required: true
+        type: number
       repo_name:
         required: true
         type: string
@@ -20,6 +23,7 @@ permissions:
   contents: read
 
 env:
+  JITTER_INDEX: ${{ inputs.jitter_index }}
   REPO: ${{ inputs.repo_name }}
   EC2_NAME: ${{ inputs.ec2_name }}
   LANGUAGE:  ${{ inputs.language }}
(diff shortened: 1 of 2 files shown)


.github/workflows/k8s-patch-os-jobs.yml
```diff
@@ -31,6 +35,10 @@ jobs:
   create-k8s-on-ec2:
     runs-on: ubuntu-latest
     steps:
+      # Sleep 20 seconds between each job's start
+      - name: Sleep to avoid throttling limits
+        run: sleep $((${{ env.JITTER_INDEX }} * 20))
```
harrryr, line 40: Maybe you could also utilize the {RANDOM} function to randomize the jitter a bit too. Not a big deal though

harrryr (approved):

.github/workflows/k8s-patch-os-jobs.yml
majanjua-amzn: Thought of that, but better to have it controlled and run in a predictable sequential manner than to randomize and potentially run into the same issue again if we get unlucky IMO.

Good callout though

PR #282, opened 2024-10-09
 Update validators for E2E runtime metrics

bjrara:
*Issue description:*
Add E2E canaries for runtime metrics in Java and Python across platforms.

*Description of changes:*
1. Code refactoring CWMetricValidator, allow verification with `ANY_VALUE` metric dimension value.
2. Update log query excluding Operation.

*Rollback procedure:*

* Java: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/11297668229/job/31425009798
* Python: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/11301593419/job/31436219251

### DIFF
--- a/validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
+++ b/validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
@@ -150,6 +150,7 @@ public class CWLogValidator implements IValidator {
     if (operation != null) {
       dependencyFilter += String.format(" && ($.Operation = \"%s\")", operation);
     } else {
+      // runtime metrics don't have Operation
       dependencyFilter += "&& ($.Operation NOT EXISTS)";
     }
 
--- a/validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
+++ b/validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
@@ -26,6 +26,7 @@ import com.amazon.aoc.services.CloudWatchService;
 import com.amazonaws.services.cloudwatch.model.Dimension;
 import com.amazonaws.services.cloudwatch.model.Metric;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 import lombok.extern.log4j.Log4j2;
 
@@ -177,20 +178,21 @@ public class CWMetricValidator implements IValidator {
          metric.getDimensions().sort(Comparator.comparing(Dimension::getName));
          actualMetricSet.add(metric);
       }
-      matchExact.removeAll(actualMetricSet);
+      Set<Metric> actualMetricSnapshot = ImmutableSet.copyOf(actualMetricSet);
+
+      actualMetricSet.removeAll(matchExact);
+      matchExact.removeAll(actualMetricSnapshot);
       if (!matchExact.isEmpty())  {
           throw new BaseException(
                   ExceptionCode.EXPECTED_METRIC_NOT_FOUND,
                   String.format(
                           "metric in %ntoBeCheckedMetricList: %s is not found in %nbaseMetricList: %s %n",
-                          matchExact.stream().findAny().get(), actualMetricSet));
+                          matchExact.stream().findAny().get(), actualMetricSnapshot));
       }
 
       Iterator<Metric> iter = matchAny.iterator();
       while (iter.hasNext()) {
           Metric expected = iter.next();
-          // There's a waste of computation because part of the actualMetricSet has been visited in the matchExact. Assuming
-          // the actualMetricSet size is small, do the comparison again to keep the logic simple.
           for (Metric actual : actualMetricSet) {
             if (metricEquals(expected, actual)) {
                 iter.remove();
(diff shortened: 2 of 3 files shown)


validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
```diff
@@ -145,7 +147,13 @@ private Map<String, Object> getActualLog(
       dependencyFilter += String.format(" && ($.RemoteResourceType = %%%s%%) && ($.RemoteResourceIdentifier = %%%s%%)", remoteResourceType, remoteResourceIdentifier);
     }
 
-    String filterPattern = String.format("{ ($.Service = %s) && ($.Operation = \"%s\") %s }", context.getServiceName(), operation, dependencyFilter);
+    if (operation != null) {
+      dependencyFilter += String.format(" && ($.Operation = \"%s\")", operation);
+    } else {
+      dependencyFilter += "&& ($.Operation NOT EXISTS)";
+    }
```
majanjua-amzn, line 154: nit: A comment explaining why we wouldn't need operation during validation would be helpful

validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
```diff
@@ -152,50 +153,57 @@ private void addMetrics(
...
+         metric.getDimensions().sort(Comparator.comparing(Dimension::getName));
+         actualMetricSet.add(metric);
+      }
+      matchExact.removeAll(actualMetricSet);
+      if (!matchExact.isEmpty())  {
+          throw new BaseException(
+                  ExceptionCode.EXPECTED_METRIC_NOT_FOUND,
+                  String.format(
+                          "metric in %ntoBeCheckedMetricList: %s is not found in %nbaseMetricList: %s %n",
+                          matchExact.stream().findAny().get(), actualMetricSet));
```
majanjua-amzn, line 186: This will print only one of potentially many missing metrics, right? It also doesn't wait until the `matchAny` check is done, could we collect all the metrics we find missing and list them for better understanding of what exactly went wrong?

validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
```diff
@@ -152,50 +153,57 @@ private void addMetrics(
...
+      }
+
+      Iterator<Metric> iter = matchAny.iterator();
+      while (iter.hasNext()) {
+          Metric expected = iter.next();
+          // There's a waste of computation because part of the actualMetricSet has been visited in the matchExact. Assuming
+          // the actualMetricSet size is small, do the comparison again to keep the logic simple.
+          for (Metric actual : actualMetricSet) {
```
majanjua-amzn, line 194: I wonder if there's a way to remove from the actual metric set the same metrics that were removed from `matchExact` during `matchExact.removeAll(actualMetricSet)` since they're already been matched.

This would reduce the number of metrics we iterate through drastically. Not a big issue though since we don't expect a huge number of metrics anyways.

validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
```diff
@@ -220,6 +228,29 @@ private List<Metric> listMetricFromCloudWatch(
...
+        return false;
+    }
+
+    private boolean dimensionEquals(Dimension expected, Dimension actual) {
+        if (expected.getName().equals(actual.getName())) {
+            return ANY_VALUE.equals(expected.getValue()) ||
+                    expected.getValue().equals(actual.getValue());
+        }
+        return false;
+    }
```
majanjua-amzn, line 252: out of scope nit: Looks like we have an opportunity to implement RegEx instead of relying on only `ANY_VALUE`. If it's not too much effort the validations could implement some regex and this could apply to all metrics instead of having to split the logic based on any value vs. exact match.

Definitely not a requirement, just a thought

majanjua-amzn (commented):
Definitely need to run this against some of our existing canaries/release tests. We can't merge this unless we're 100% sure it will not break our log and metric validation.

Unit tests don't provide enough confidence here, even if they have good coverage, since the real data has many different metric aggregations that can be impacted differently.

validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
bjrara: Yes, this is how it works today.

We are able to identify all missing metrics because the deduplicated `actualMetrics` are all printed out. Since we know the template to compare, we can figure out the diff.

validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
bjrara: Supporting regex means we can no longer utilize hash set to remove the exact match metrics quickly. It may bring performance concern because the algorithm complexity will increase.

validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
bjrara: Added

validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
bjrara: Done.

harrryr (approved):
LGTM

PR #310, opened 2024-11-02
Set dotnet E2E nuget test target for dotnet sdk v1.3.2

AsakerMohd:
*Description of changes:*
Set current .NET SDK release for .NET E2E Nuget instead of the latest version because the new ADOT .NET SDK release will introduce breaking changes. This will fix the current Nuget E2E failure.


vastin (approved):
LGTM

PR #339, opened 2024-12-24
Add lambda layer resource cleanup.

zzhlogin:
*Description of changes:*
Add lambda layer resource cleanup script.

*Rollback procedure:*
Revert the commit.

Test workflow passed: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/12485707531/job/34844864812


.github/workflows/util/clean/lambda_layer_cleanup/cleaner.py
```diff
@@ -0,0 +1,61 @@
...
+            # Ensure `created_date` is timezone-aware
+            created_date = created_date.astimezone(timezone.utc)
+
+            if created_date < cutoff_time:
+                api_id = api['id']
+                api_name = api.get('name', 'Unnamed API')
+                print(f"Preparing to delete API: {api_name} (ID: {api_id}), created at {created_date}")
+
+                # Attempt to delete the API with retries
+                delete_api_with_retries(client, api_id)
```
majanjua-amzn, line 49: We should always sleep >=30 seconds after deleting. There is a rate limit of 1 request per 30 seconds for the delete API.
https://docs.aws.amazon.com/apigateway/latest/developerguide/limits.html

I guess you can play with a default sleep + exponential back-off combination here

.github/workflows/util/clean/lambda_layer_cleanup/cleaner.py
```diff
@@ -0,0 +1,61 @@
...
+    for api in apis.get('items', []):
+        created_date = api.get('createdDate')  # This is usually UTC already
+        if created_date and isinstance(created_date, datetime):
+            # Ensure `created_date` is timezone-aware
+            created_date = created_date.astimezone(timezone.utc)
+
+            if created_date < cutoff_time:
+                api_id = api['id']
+                api_name = api.get('name', 'Unnamed API')
+                print(f"Preparing to delete API: {api_name} (ID: {api_id}), created at {created_date}")
```
majanjua-amzn, line 46: Are we sure there aren't any other APIGW REST APIs that are needed for other tests and persist in the accounts? I think we should parse by name as well to make this future proof in case it does happen.

An if statement to check if the name of the APIGW is of the correct structure, e.g. `AdotLambdaPython...`.

majanjua-amzn (approved):

PR #340, opened 2024-12-25
Final code for dotnet sample app deployment

Jeel-mehta:
*Issue description:*

*Description of changes:* I have dynamically deployed the sample app for both the versions of dotnet support (8.0 and 6.0). I am attaching a link of github workflow where I have tested my code for version 6.0 and 8.0 for the region us-east-1. This specific code change will include and handle both versions for all the regions. 

*Rollback procedure:*


.github/workflows/dotnet-sample-app-s3-deploy.yml
```diff
@@ -42,13 +46,68 @@ jobs:
...
+            grep -n "TargetFramework" asp_remote_service/asp_remote_service.csproj
+
+          else
+            echo "Using default version ${DEFAULT_VERSION}, no changes needed"
+          fi
+
       - name: Build Sample App Zip
         working-directory: sample-apps/dotnet
-        run: zip -r dotnet-sample-app.zip .
+        run: zip -r dotnet-sample-app:${DOTNET_VERSION}.zip .
```
majanjua-amzn, line 106: nit: Shouldn't use `:` in file name
```suggestion
        run: zip -r dotnet-sample-app-${DOTNET_VERSION}.zip .
```

.github/workflows/dotnet-sample-app-s3-deploy.yml
```diff
@@ -42,13 +46,68 @@ jobs:
...
 
       - name: Upload to S3
         working-directory: sample-apps/dotnet
-        run: aws s3api put-object --bucket aws-appsignals-sample-app-prod-${{ matrix.aws-region }} --body ./dotnet-sample-app.zip --key dotnet-sample-app.zip
+        run: aws s3api put-object --bucket aws-appsignals-sample-app-prod-${{ matrix.aws-region }} --body ./dotnet-sample-app:${DOTNET_VERSION}.zip --key dotnet-sample-app:${DOTNET_VERSION}.zip
```
majanjua-amzn, line 110:
```suggestion
        run: aws s3api put-object --bucket aws-appsignals-sample-app-prod-${{ matrix.aws-region }} --body ./dotnet-sample-app-${DOTNET_VERSION}.zip --key dotnet-sample-app-${DOTNET_VERSION}.zip
```

majanjua-amzn (changes requested):
Instead of looking for the default version `'8.0'` and replacing it, you should change wherever it says `8.0` to `{{DOTNET_VERSION}}` and use sed commands to replace all of the occurrences of that string

.github/workflows/dotnet-sample-app-s3-deploy.yml
harrryr: `:` is for tags in images. For file name `-` is sufficient

.github/workflows/dotnet-sample-app-s3-deploy.yml
```diff
@@ -42,13 +46,68 @@ jobs:
...
+              if [ ! -f "$file" ]; then
+                echo "Error: File $file not found"
+                exit 1
+              fi
+              
+              # Update TargetFramework in csproj files
+              if [[ $file == *.csproj ]]; then
+                sed -i "s/<TargetFramework>netcoreapp${DEFAULT_VERSION}<\/TargetFramework>/<TargetFramework>netcoreapp${DOTNET_VERSION}<\/TargetFramework>/g" "$file"
+              # Update version in Dockerfiles
+              elif [[ $file == */Dockerfile ]]; then
```
harrryr, line 67: I don't believe ec2 case uses the Dockerfile. We can remove this condition and the content inside it

harrryr: It will be useful for the EKS case though if we end up implementing it

Jeel-mehta: Hey, Harry should I retain the Dockerfile changes or should I just delete it?

.github/workflows/dotnet-sample-app-s3-deploy.yml
Jeel-mehta: Okay, I will make changes and change from : to - wherever it is needed

.github/workflows/dotnet-sample-app-s3-deploy.yml
harrryr: Delete it from this PR

Jeel-mehta: Only the docker part right?

majanjua-amzn: Remove the condition here too, it doesn't make sense to have it here

Jeel-mehta: Okay, I will remove the docker part

.github/workflows/dotnet-sample-app-s3-deploy.yml
```diff
@@ -42,13 +46,55 @@ jobs:
           role-to-assume: arn:aws:iam::${{ env.ACCOUNT_ID }}:role/${{ secrets.E2E_TEST_ROLE_ARN }}
           aws-region: ${{ matrix.aws-region }}
 
+      - name: Update .NET version if not default
+        working-directory: sample-apps/dotnet
+        run: |
+          if [ "${DOTNET_VERSION}" != "${DEFAULT_VERSION}" ]; then
```
harrryr, line 52: Since we removed 8.0 and replaced it with `{DEFAULT_VERSION}`, the sed command should always be running so this if statement should not be here

.github/workflows/dotnet-sample-app-s3-deploy.yml
```diff
@@ -42,13 +46,55 @@ jobs:
...
+            update_version() {
+              local file=$1
+              if [ ! -f "$file" ]; then
+                echo "Error: File $file not found"
+                exit 1
+              fi
+              
+              # Update TargetFramework in csproj files
+              if [[ $file == *.csproj ]]; then
+                sed -i "s/<TargetFramework>netcoreapp${DEFAULT_VERSION}<\/TargetFramework>/<TargetFramework>netcoreapp${DOTNET_VERSION}<\/TargetFramework>/g" "$file"
```
harrryr, line 65: A large portion of the string is redundant. Just look for ${DOTNET_VERSION}

```suggestion
                sed -i "s/${DEFAULT_VERSION}/${DOTNET_VERSION}/g" "$file"
```

harrryr (approved):

Jeel-mehta:
> requested change

Hey Mahad, I think if we have a dynamic variable {{DOTNET_VERSION}} instead of 8.0, we are just handling the csproj files currently(frontend and remote service). Right now we are not considering Dockerfiles, but since Dockerfiles and csproj files are important when it comes to eks clusters at that time they won't be uniform. Also while working with eks clusters, the Dockerfiles call csproj files and since they will not be uniform, it will throw an error.

.github/workflows/dotnet-sample-app-s3-deploy.yml
Jeel-mehta: Hey Harry, currently what you suggested works but in the future
This command would replace ALL instances of the version number anywhere in the file, which could be problematic because:
It would replace the version number even if it appears in comments
It would replace the version number in any other XML elements or attributes
It might replace version numbers that shouldn't be changed
It could accidentally replace partial matches (for example, if DEFAULT_VERSION is "6.0" and the file contains "16.0" somewhere), so is it okay to not change this?

majanjua-amzn (approved):

PR #345, opened 2025-01-03
Fix Lambda canary trace validation failure

wangzlei:
*Issue description:*
When searching the trace by API GetTraceSummaries, if the end time of searching is set to be far away from the actual trace end time, xray does not return Lambda traces. Right now I am not sure it is an aws sdk xray client issue or xray service issue. This PR is a workaround by moving trace validation step to be earlier, so the endTime of searching, which the value is the current time, is close to the actual trace end time. Then xray service can return the Lambda trace and pass the trace validation.

*Description of changes:*

*Rollback procedure:*

### DIFF
--- a/.github/workflows/node-lambda-test.yml
+++ b/.github/workflows/node-lambda-test.yml
@@ -148,6 +148,21 @@ jobs:
         shell: bash
         run: sleep 30s; curl -sS ${{ env.API_GATEWAY_URL }}
 
+      - name: Validate generated traces
+        id: trace-validation
+        if: (success() || steps.log-validation.outcome == 'failure' || steps.metric-validation.outcome == 'failure') && !cancelled()
+        # will be removed after data quality bug fixed
+        continue-on-error: true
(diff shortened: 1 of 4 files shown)


.github/workflows/node-lambda-test.yml
```diff
@@ -148,12 +148,12 @@ jobs:
         shell: bash
         run: sleep 30s; curl -sS ${{ env.API_GATEWAY_URL }}
 
-      # Validation for pulse telemetry data
-      - name: Validate generated EMF logs
-        id: log-validation
+      - name: Validate generated traces
+        id: trace-validation
+        if: (success() || steps.log-validation.outcome == 'failure' || steps.metric-validation.outcome == 'failure') && !cancelled()
```
majanjua-amzn, line 153: `if` conditions of all the validation steps need to be changed to account for the new order. The purpose of the `if` conditions being to ensure that each type of validation runs even if the others were failed

wangzlei: update the "if" for 3 validations

majanjua-amzn (approved):

PR #369, opened 2025-02-14
Rename service name

pxaws:
*Issue description:*

*Description of changes:*

Shorten the remote deployment name so that the relevant pod name follow the regular regex pattern. See details here: https://github.com/aws/amazon-cloudwatch-agent/pull/1553

*Rollback procedure:*

* [Test Java EKS](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/13336222797/job/37251919576)
* [Test Python EKS](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/13335460362)
* [Test Node EKS](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/13335591495/job/37251408435)
* [Test Dotnet EKS](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/13336222797/job/37251919576)


terraform/java/eks/main.tf
```diff
@@ -205,11 +205,14 @@ resource "kubernetes_deployment" "sample_remote_app_deployment" {
...
-resource "kubernetes_service" "sample_remote_app_service" {
+resource "kubernetes_service" "sample_remote_app_deployment" {
   depends_on = [ kubernetes_deployment.sample_remote_app_deployment ]
 
   metadata {
-    name = "sample-remote-app-service"
+    # use the same name as the deployment to handle the edge case when the deployment name is longer than 47 characters
+    # in this edge case, we just use the service name (rather than deployment name) as RemoteService
+    # see https://github.com/aws/amazon-cloudwatch-agent/pull/1553
+    name      = "sample-r-app-deployment-${var.test_id}"
```
majanjua-amzn, line 215: Is it expected of customers to have the same service and deployment name when the names are long? Or in all cases? This seems unlikely and a bad expectation to have

majanjua-amzn (commented):
I would prefer if we can still identify that this value refers to the deployment.

I would also prefer that we validate the resulting behaviour, not tailor the tests to get the telemetry we want. This feels backwards, instead of updating the setup we should be testing that having a long name results in the correct service name. We are simply avoiding the issue with this PR and we lose visibility.

pxaws:
> I would prefer if we can still identify that this value refers to the deployment.
> 
> I would also prefer that we validate the resulting behaviour, not tailor the tests to get the telemetry we want. This feels backwards, instead of updating the setup we should be testing that having a long name results in the correct service name. We are simply avoiding the issue with this PR and we lose visibility.

We have python, dotnet, node tests still use the shorter deployment. In this case, the Remote Service will take the value of the deployment. So they are still testing the old behavior as before. For the java test, we choose to keep the long deployment name and set the service name to be the same as deployment name (to test the edge case where we will use the service name if pod name doesn't follow normal regex pattern).

terraform/java/eks/main.tf
pxaws: Yes, we will suggest the workaround to customers. We can't cover all the edge cases. This is just the implication of choosing to use the endpoint slice API as we don't have enough information for pod to get the corresponding deployment. Another way is to let customers to use the environmental variable to switch to the old behavior if it's really needed. 

Currently we have two options: 

1. keep the old behavior which means we will break customers with big cluster (with a few thousands nodes and tens of thousands pods) 
2. switch to the new behavior which has limitations but can be worked around 

To me, option 2 is clearly the way to go.

majanjua-amzn (approved):

PR #377, opened 2025-02-28
[Python] Add ADOT SigV4 release test workflow + validations

majanjua-amzn:
## Background

ADOT instrumentation has the newly added AWS SigV4 OTLP exporters, for which we need to add release tests. These tests serve to ensure functionality is not changing over time. The new workflow will be called from the aws-otel-python-instrumentation repository.

## Changes

- New test that runs in us-west-2 of the us-east-1 test account (this is because transaction search needs to be enabled in a region that is not already being used for other testing)
- New validations
- Added logic in the `CWLogValidator` to get OTLP span logs when applicable

### Follow up
Need to follow up this PR with one in aws-otel-python-instrumentation to add the test to the main build workflow

## Rollback Procedure

Revert

## Testing

Ran the workflow and another test to ensure no adverse effect on the canaries: [link](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/13578588516)

---

### DIFF
--- a/validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
+++ b/validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
@@ -71,8 +71,26 @@ public class CWLogValidator implements IValidator {
           String remoteResourceType = (String) expectedAttributes.get("RemoteResourceType");
           String remoteResourceIdentifier = (String) expectedAttributes.get("RemoteResourceIdentifier");
 
-          Map<String, Object> actualLog =
-                  this.getActualLog(operation, remoteService, remoteOperation, remoteResourceType, remoteResourceIdentifier);
+          Map<String, Object> actualLog;
+
+          // Parsing unique identifiers in OTLP spans
+          if (operation == null) {
+            operation = (String) expectedAttributes.get("attributes[\\\"aws.local.operation\\\"]");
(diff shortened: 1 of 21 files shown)


validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
```diff
@@ -71,8 +71,26 @@ public void validate() throws Exception {
...
 
-          Map<String, Object> actualLog =
-                  this.getActualLog(operation, remoteService, remoteOperation, remoteResourceType, remoteResourceIdentifier);
+          Map<String, Object> actualLog;
+
+          // Parsing unique identifiers in OTLP spans
+          if (operation == null) {
+            operation = (String) expectedAttributes.get("attributes[\\\"aws.local.operation\\\"]");
+            remoteService = (String) expectedAttributes.get("attributes[\\\"aws.remote.service\\\"]");
+            remoteOperation = (String) expectedAttributes.get("attributes[\\\"aws.remote.operation\\\"]");
```
harrryr, line 80: Looks like we changed from getting the remoteService and remoteOperation from `RemoteOperation` to `aws.remote.service`. 

Will this not affect the query for the existing `this.getActualLog` method?

harrryr: Nvm, looks like we are extending the coverage for getActualLog too

harrryr (approved):

PR #391, opened 2025-04-09
[Python] Add Performance Testing Workflow for Lambda Layer

liustve:
**Issue description:**
Added a workflow to automatically performance test and gather the results for the unreleased Python Lambda Layer.

**The GitHub workflow does the following:**
1. Deploy and setup the Python Lambda sample application instrumented with the Application Signals Lambda Layer on the AppSignals e2e testing aws account.
2. Record the start time of the test and run the specified amount (default: 20) Lambda cold start iterations on the sample application
3. Record the end time of the test
4. Query CW Logs Insights for the aggregated performance testing metrics with the following query:
```
fields @timestamp, @message, @initDuration
| filter @message like "REPORT RequestId"
| stats
    count(@initDuration) as Sample_Count,
    avg(@initDuration) as Average_Init_Duration,
    min(@initDuration) as Min_Init_Duration,
    max(@initDuration) as Max_Init_Duration,
    pct(@initDuration, 50) as P50_Init_Duration,
    pct(@initDuration, 90) as P90_Init_Duration,
    pct(@initDuration, 99) as P99_Init_Duration'
```
5. Parse and flatten the result from the query call and save it in a file.
6. Clean and destroy the deployed lambda resources upon completion, cancellation, or failure.

**Testing**
Testing was done on a personal aws dev account with the exact IAM permissions as the e2e testing aws account.
Example of a successful job:

https://github.com/liustve/aws-application-signals-test-framework/actions/runs/14367621556/job/40284171797


lambda-layer-perf-test/lambda-layer-run.sh
```diff
@@ -0,0 +1,110 @@
...
+  else
+      ENV_JSON="{\"Variables\":{\"AWS_LAMBDA_EXEC_WRAPPER\":\"/opt/otel-instrument\",\"FOO\":\"BAR_$i\"}}"
+  fi
+
+  echo "Iteration $i: Updating environment variables to simulate cold start"
+
+  # Update environment variables
+  aws lambda update-function-configuration \
+    --function-name $FUNCTION_NAME \
+    --environment "$ENV_JSON" > /dev/null
```
yiyuan-he, line 33: Where do we set up this Lambda? Do we use a pre-existing one?

lambda-layer-perf-test/lambda-layer-run.sh
```diff
@@ -0,0 +1,110 @@
...
+
+  # Wait a short time for changes to take effect
+  sleep 5
+
+  # Invoke the Lambda function
+  echo "Iteration $i: Invoking Lambda function"
+  aws lambda invoke --function-name $FUNCTION_NAME --payload '{}' response.json > /dev/null
```
yiyuan-he, line 40: What do we do if invoke Lambda fails?

.github/workflows/python-lambda-layer-perf-test.yml
```diff
@@ -0,0 +1,108 @@
...
+        name: no-layer-results
+        path: no_layer_results.txt
+        retention-days: 90
+          
+    - name: Upload layer results
+      uses: actions/upload-artifact@v4
+      with:
+        name: layer-results
+        path: layer_results.txt
+        retention-days: 90
```
yiyuan-he, line 100: How is user expected to retrieve results? 

Currently from [the workflow](https://github.com/liustve/aws-application-signals-test-framework/actions/runs/14367621556/job/40284171797), I see I have to click into different stages and download multiple zip files. Can we consolidate to one artifact?

lambda-layer-perf-test/lambda-layer-run.sh
```diff
@@ -0,0 +1,110 @@
+## Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+## SPDX-License-Identifier: Apache-2.0
+
+#!/bin/bash
+
+IS_BASE_RUN=${1:-false}
+FUNCTION_NAME=${2:-none}
+SLEEP_TIME_SECONDS=300
```
yiyuan-he, line 8: Where does this number come from? Can we guaranteee the CloudWatch logs will always appear in this timeframe?

lambda-layer-perf-test/lambda-layer-run.sh
```diff
@@ -0,0 +1,110 @@
...
+  # Wait a short time for changes to take effect
+  sleep 5
+
+  # Invoke the Lambda function
+  echo "Iteration $i: Invoking Lambda function"
+  aws lambda invoke --function-name $FUNCTION_NAME --payload '{}' response.json > /dev/null
+
+  # Wait to simulate cold start reset
+  echo "Iteration $i: Waiting for cold start reset"
+  sleep 5
```
yiyuan-he, line 44: Does 5 sec always produce cold start? If there happens to be no cold start, how can we mitigate skewed averages in data?

lambda-layer-perf-test/lambda-layer-run.sh
liustve: This is being done in the ./build.sh script:
https://github.com/aws-observability/aws-otel-python-instrumentation/blob/main/lambda-layer/build.sh

lambda-layer-perf-test/lambda-layer-run.sh
liustve: Good point, should we have completely fail the run in that case and ask the person running it to restart the workflow?

.github/workflows/python-lambda-layer-perf-test.yml
liustve: I can consolidate it but the job uploads the artifact under the run:
https://github.com/liustve/aws-application-signals-test-framework/actions/runs/14367621556

wangzlei (approved):

lambda-layer-perf-test/lambda-layer-run.sh
liustve: Varying amounts of testing done, I can't guarantee that all runs will be present in the logs when I add this delay but so far it's worked.

liustve: There was a bit of flakiness if I add a smaller delay (i.e. 1.5 - 2 minutes) where the number of sampled results returned from the query would give 18 or 19 instead of the expected 20 runs.

lambda-layer-perf-test/lambda-layer-run.sh
liustve: Maybe should update the preflight steps to ask the person to cancel and restart the workflow if the invoke call fails?

.github/workflows/python-lambda-layer-perf-test.yml
yiyuan-he: If it's not too difficult, I think one artifact would be nice.

Consider when we add this automation for each language: That will be 8 artifacts to download. In the future, when we support even more languages it may get unmanageable especially since we usually do lambda layer releases all at once.

lambda-layer-perf-test/lambda-layer-run.sh
yiyuan-he: Sounds good. We can err on the side of caution and add extra buffer if necessary.

Another thing to consider is: What happens when we add this automation for all languages and the tests are run closely together? Are we able to partition the metrics from each language correctly?

lambda-layer-perf-test/lambda-layer-run.sh
yiyuan-he: Yeah that is fine as long as we document it somewhere in MCM.

yiyuan-he:
Can you trigger a workflow re-run to view the new artifact

liustve:
> Can you trigger a workflow re-run to view the new artifact

Verified in nodejs run with same code:
https://github.com/liustve/aws-application-signals-test-framework/actions/runs/14368905758

yiyuan-he (approved):
LGTM thanks!

PR #425, opened 2025-07-09
[Genesis] Release Testing - Validator Changes

liustve:
*Description of changes:*

This PR is part 1 of the release testing changes to support the incoming release of the latest Genesis/Agentic AI Observability changes done in ADOT Python which contains all of the changes to `validator`

We need to validate that we are correctly emitting all 3 telemetry signals to CW backend.

1. Expands `Context` class to also take in a given traceId - as a part of the Agent Observability, we've split off user input/output prompts from Gen AI span events as a separate log event. The idea is that users will be able to rebuild their entire Gen AI span event by querying for logs with the same trace id in `aws/spans` and a custom log group. As a part of this testing we need to verify that trace Id is being propagated from their `Agentic AI application + ADOT --> CW` correctly and can be linked together.

2. Updated `AWS OTLP log filter` patterns to capture and validate GenAI input/output prompt log events.

3.  Updated `TraceValidator` to validate pre-generated trace IDs for GenAI release testing

4. Updated `MetricValidator` to handle GenAI metrics special case: ADOT Python captures all OTel metrics from the Agentic AI application Instrumentor and converts them to EMF format. Validation for this path just checks for any metric emission to a custom namespace since the exact metric structure is unknown

Example of a successful run of this test: https://github.com/liustve/aws-application-signals-test-framework/actions/runs/16172487807/job/45649328445

### DIFF
--- a/validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
+++ b/validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
@@ -147,15 +147,15 @@ public class TraceValidator implements IValidator {
               validationConfig.getHttpMethod().toUpperCase(),
               validationConfig.getHttpPath()));
     }
+
+    if (validationConfig.getHttpPath().contains("ai-chat")) {
+      return this.getTraceById(Collections.singletonList(context.getTraceId()));
+    }
+
     log.info("Trace Filter: {}", traceFilter);
     List<TraceSummary> retrieveTraceLists = xrayService.searchTraces(traceFilter);
     List<String> traceIdLists = Collections.singletonList(retrieveTraceLists.get(0).getId());
-    List<Trace> retrievedTraceList = xrayService.listTraceByIds(traceIdLists);
-
-    if (retrievedTraceList == null || retrievedTraceList.isEmpty()) {
-      throw new BaseException(ExceptionCode.EMPTY_LIST);
-    }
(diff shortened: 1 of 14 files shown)


validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
```diff
@@ -147,15 +147,15 @@ private Map<String, Object> getTrace() throws Exception {
               validationConfig.getHttpMethod().toUpperCase(),
               validationConfig.getHttpPath()));
     }
+
+    if (validationConfig.getHttpPath().contains("ai-chat")) {
```
majanjua-amzn, line 151: Do we need to do a null check here?
```suggestion
    if (validationConfig.getHttpPath() != null && validationConfig.getHttpPath().contains("ai-chat")) {
```

liustve: Yes, if we do `validationConfig.getHttpPath().contains("ai-chat")` then this will throw a NullPointerException as `validationConfig.getHttpPath()` will return null. This was causing one of the unit tests to break:
https://github.com/aws-observability/aws-application-signals-test-framework/pull/425/files#diff-0e227ac57effe54c79f6c9fcb39a24269103137e57dd4acd8d711b07d47ffe90R158

PR #628, opened 2026-05-15
fix: add ECS cleanup to prevent orphaned resources from cancelled workflows

liustve:
- Adds a new ECS cluster cleanup job to the daily `resource-cleanup.yml` workflow that deletes orphaned ECS clusters/services older than 3 hours with the `e2e-test-` prefix
- Adds a fallback force-cleanup step to `node-ecs-test.yml` that uses AWS CLI to delete ECS resources if Terraform destroy fails or times out

When a workflow is cancelled mid-flight (e.g., due to `concurrency: cancel-in-progress: true`), the `terraform destroy` step may not execute. This leaves orphaned ECS clusters/services running indefinitely, triggering Mirador container vulnerability findings.

example run: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/25894682424/job/76105171049


.github/workflows/resource-cleanup.yml
```diff
@@ -81,13 +81,36 @@ jobs:
...
+          python -m pip install -r requirements.txt
+          python cleaner.py
+
   cleanup-lambda-layer:
     strategy:
       fail-fast: false
       matrix:
         aws-region: ['af-south-1','ap-east-1','ap-northeast-1','ap-northeast-2','ap-northeast-3','ap-south-1','ap-south-2','ap-southeast-1',
                      'ap-southeast-2','ap-southeast-3','ap-southeast-4','ca-central-1','eu-central-1','eu-central-2','eu-north-1',
-                     'eu-south-1','eu-south-2','eu-west-1','eu-west-2','eu-west-3','il-central-1','me-central-1','me-south-1', 'sa-east-1',
```
jj22ee, line 90: Why skip `me-south-1` here?

.github/workflows/resource-cleanup.yml
```diff
@@ -81,13 +81,36 @@ jobs:
           python -m pip install -r requirements.txt
           python cleaner.py
 
+  cleanup-ecs-clusters:
```
jj22ee, line 84: Since we have this, then we don't need the changes to `<languages>-ecs-test.yml`?

I think it's simpler to just only have this, otherwise we're in a state where we have extra cleanup scripts that are only present in 3 out of the many other test workflows.

.github/workflows/resource-cleanup.yml
liustve: me-south-1 is completely down so the workflow has been failing

.github/workflows/resource-cleanup.yml
liustve: done!

ezhang6811 (approved):

PR #629, opened 2026-05-27
fix: use prefix-based arg matching for k8s operator image patch

wangzlei:
## Summary
- K8s deploy terraform scripts used hardcoded arg indices (`args/0`, `args/2`, etc.) to patch the auto-instrumentation image on the CW Agent Operator deployment. When the operator added new args (e.g. `--auto-monitor-config`), the indices shifted and the patch overwrote the wrong arg, leaving the actual instrumentation image unchanged.
- This caused k8s tests to silently use the operator's default public image instead of the CI-built staging image, meaning the tests were never actually validating the latest code from the SDK repos.
- Replace the hardcoded JSON patch with the same `jq map/test` approach already used by the EKS `patch_image_and_check_diff` action, which matches args by prefix and is resilient to ordering changes.
- Update `java/k8s` validation templates to expect `www.amazon.com:443` (with port), matching the behavior of the current ADOT Java agent and aligning with the existing `java/eks` templates.

## Test plan
- [ ] Run java-k8s test and verify the staging SNAPSHOT image is actually injected (check `Telemetry.SDK` contains `-SNAPSHOT`)
- [ ] Verify log/metric/trace validations pass with the updated templates
- [ ] Confirm EKS tests are unaffected

### DIFF
--- a/terraform/dotnet/k8s/deploy/main.tf
+++ b/terraform/dotnet/k8s/deploy/main.tf
@@ -91,8 +91,10 @@ resource "null_resource" "deploy" {
         sleep 10
         kubectl wait --for=condition=Ready pod --all -n amazon-cloudwatch
       elif [ "${var.repository}" = "aws-otel-dotnet-instrumentation" ]; then
-        kubectl patch deploy -n amazon-cloudwatch amazon-cloudwatch-observability-controller-manager --type='json' \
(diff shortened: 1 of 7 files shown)


v-iw (approved):
good

ezhang6811 (approved):

PR #631, opened 2026-05-27
Revert k8s terraform operator patch changes from #629 and #630

wangzlei:
## Summary
- Reverts the terraform changes from #629 (prefix-based arg matching) and #630 (timeout fix) for all 4 language k8s deploy scripts.
- The patch correctly updates the operator's `--auto-instrumentation-*-image` arg, but the k8s EC2 node cannot pull the private staging image from `611364707713.dkr.ecr.us-west-2.amazonaws.com` without proper ECR authentication, causing sample app pods to remain in ImagePullBackOff.
- Restores the previous behavior where k8s tests use the operator's default public image until the ECR pull credential issue is resolved.
- The validation template updates (`www.amazon.com:443`) from #629 are intentionally kept, as those will be needed when we re-enable the patch with proper credentials.

## Test plan
- [ ] Run java-k8s test and verify it passes with operator default image
- [ ] Verify EKS tests are unaffected (they use a separate patch mechanism)


v-iw (approved):
good

PR #647, opened 2026-06-19
Add Python EC2 Service Events E2E test + PromQL validator

vastin:
## What

Adds the **Python EC2 Service Events** E2E test cell and the **PromQL validator** it depends on.

Service Events is bundled with Application Signals: enabling App Signals turns it on (Lambda excluded). The instrumented Django sample app then emits five signals autonomously, validated end-to-end against CloudWatch:

| Signal | Type | How it's triggered |
|---|---|---|
| `DeploymentEvent` | OTLP log (`/aws/service-events/<svc>`) | emitted on startup |
| `IncidentSnapshot` (exception) | OTLP log (same group) | `/mysql` 500 + captured exception |
| `IncidentSnapshot` (latency) | OTLP log (same group) | healthy 200 route exceeding `LATENCY_THRESHOLDS` |
| `EndpointErrorMetric` `count` | OTLP metric (PromQL) | the `/mysql` 500 |
| `FunctionCall` `service.function.duration` | OTLP metric (PromQL) | scoped via `PACKAGES_INCLUDE` |

The two log signals use the existing cw-log validator. The metric signals are native OTLP metrics with no fixed CloudWatch namespace, so they are validated with a new **PromQL validator** against the CloudWatch Prometheus-compatible query API (`monitoring.<region>.amazonaws.com`).

## Components

- `validator/` — new `PromQLValidator` + `PromQLService` (SigV4-signed query), wired into `ValidatorFactory` / `PredefinedExpectedTemplate`.
- `terraform/python/ec2/service-events/` — single-instance module: instrumented Django frontend + traffic generator, CW Agent (App Signals config only, which provisions the 4316 OTLP receiver + Service Events routing).
- `.github/workflows/python-ec2-service-events-test.yml` — reusable workflow (callable + dispatchable), matrixable over Python versions.
- Expected-data templates + validation YAMLs for all five signals.

## Validation

Validated end-to-end in CI against the org E2E account via workflow_dispatch run [`28046120624`](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/28046120624) — completed `success` with all five validator steps (DeploymentEvent, both IncidentSnapshots, EndpointErrorMetric, FunctionCall) reporting `success`, terraform deploy + destroy clean. Run on the final rebased branch (rebased onto `main` past #648/#649; the unused `canary_type` variable was dropped to match #649).

## Wheel source

The workflow installs the released ADOT distro via the standard `releases/latest` canary pattern. Service Events shipped in `aws-opentelemetry-distro` **v0.18.0**, so the released wheel now carries the `serviceevents` code and emits signals. (Earlier revisions installed from git `main` as a temporary measure while the feature was main-only; that has been reverted.)


validator/src/main/resources/expected-data-template/python/ec2/service-events/incident-snapshot.mustache
```diff
@@ -0,0 +1,37 @@
...
+    "vcs.repository.url.full": "^https?://.+$",
+    "http.request.method": "^GET$",
+    "http.response.status_code": "^500$",
+    "url.route": "^mysql$"
+  },
+  "body": {
+    "request_context": {
+      "status_code": "^500$",
+      "type": "^http$"
+    }
```
jj22ee: Should we also validate the existence of `body.exception_info` and its content in the `incident-snapshot` (and `incident-snapshot-latency`) validation?

```
"body": {
  "request_context": { "status_code": "^500$", "type": "^http$" },
  "exception_info": [ { "exception_type": "^KeyError$", "call_path": [ { "function_name": "^.+$" } ] } ]
}
```

validator/src/main/resources/expected-data-template/python/ec2/service-events/incident-snapshot.mustache
```diff
@@ -0,0 +1,37 @@
...
+      "aws.service_events.deployment.id": "{{serviceName}}",
+      "deployment.environment.name": "^ec2:service-events$",
+      "cloud.platform": "^aws_ec2$"
+    }
+  },
+  "scope": {
+    "name": "serviceevents"
+  },
+  "eventName": "^aws.service_events.incident_snapshot$",
+  "traceId": "^([a-f0-9]{32})?$",
```
jj22ee, line 14: I recall that in real service-events data, I saw that `traceId` is empty string, which I guess is why the validation here uses `?` to make the value validation optional. Wondering if we have planned to always try to populate the value when possible in the SDKs?

validator/src/main/resources/expected-data-template/python/ec2/service-events/function-call-metric.mustache
```diff
@@ -0,0 +1,23 @@
+[
+  {
+    "Telemetry.Source": "^ServiceEvents$",
+    "__name__": "^service.function.duration$",
```
jj22ee, line 4: Can we also validate/assert metric values? Maybe check for `> 0` for function calls in `PromQLValidator`. Also same comment for `endpoint-error-metric`.

validator/src/main/java/com/amazon/aoc/validators/PromQLValidator.java
```diff
@@ -0,0 +1,177 @@
...
+ * regex-per-field convention of {@link CWLogValidator}.
+ *
+ * <p>The PromQL query itself lives in {@code validationConfig.promQlQuery} and is the only place the
+ * metric name is pinned. CloudWatch's PromQL surface exposes OTLP datapoint attributes either bare
+ * ({@code operation}) or prefixed ({@code @datapoint.operation}); a single label key in the
+ * expected template is therefore matched against both the bare key and its {@code @datapoint.}
+ * prefixed form so the test does not break on either representation.
+ */
+@Log4j2
+public class PromQLValidator implements IValidator {
```
jj22ee, line 55: Since we're updating the validator in `validator/src/`, we should add unit tests in `validator/src/test/...` for PromQL.

validator/src/main/resources/expected-data-template/python/ec2/service-events/incident-snapshot.mustache
vastin: On the Python/Django path the IncidentSnapshot intermittently emits an empty traceId/spanId. The snapshot is occasionally assembled without an active span context attached, so the field comes through blank. There will be another PR to refactor python hooks into span processor and it will fix the issue. Once it is fix, the e2e validation logic will be updated.

validator/src/main/resources/expected-data-template/python/ec2/service-events/incident-snapshot.mustache
vastin: fixed.

validator/src/main/resources/expected-data-template/python/ec2/service-events/function-call-metric.mustache
vastin: Added

validator/src/main/java/com/amazon/aoc/validators/PromQLValidator.java
vastin: Added

validator/src/main/resources/expected-data-template/python/ec2/service-events/deployment-event.mustache
```diff
@@ -0,0 +1,21 @@
...
+  "scope": {
+    "name": "serviceevents"
+  },
+  "eventName": "^aws.service_events.deployment_event$",
+  "attributes": {
+    "event.name": "aws.service_events.deployment_event",
+    "aws.service_events.deployment.trigger": "^(startup|periodic|shutdown)$",
+    "aws.service_events.deployment.id": "{{serviceName}}",
+    "vcs.ref.head.revision": "^[0-9a-f]{40}$",
+    "vcs.repository.url.full": "^https?://.+$"
```
jj22ee, line 19: Can also consider validating the exact values.

jj22ee (approved):

validator/src/main/resources/expected-data-template/python/ec2/service-events/deployment-event.mustache
vastin: Sure. it will be addressed in follow-up PR.

PR #655, opened 2026-06-25
tests: adding Otel lite sdk e2e tests

wespil:
# Add Lambda Lite SDK (Fast Start) e2e and performance tests for Python and Node.js

## Summary

Adds end-to-end and cold-start performance test coverage for the ADOT Lambda
**Lite SDK / Fast Start** mode (`OTEL_AWS_LAMBDA_FAST_START=true`) for both
Python and Node.js, mirroring the existing non-lite Lambda test setup. Also
introduces a single combined performance script that exercises all three
cold-start scenarios against one function.

### E2E test workflows (reusable, `workflow_call`)
- `python-lambda-lite-sdk-test.yml` — deploy → invoke → validate traces → destroy
- `node-lambda-lite-sdk-test.yml` — same flow for Node.js
- Both check out the `otel-lite-sdk` branch when invoked from the `main-build`
  caller so they validate against this branch's terraform/templates.

### Performance test workflows (`workflow_dispatch`)
- `python-lambda-lite-sdk-perf-test.yml`
- `node-lambda-lite-sdk-perf-test.yml` — runs the combined script across
  Full SDK, Lite SDK (fast start), and No Layer (baseline).

### Combined performance script
- `lambda-layer-perf-test/lambda-perf-test-all.sh` — runs all three cold-start
  scenarios against the **same** Lambda function in one invocation. Each
  scenario records its own CloudWatch time window, so a single trailing
  log-ingestion sleep brackets every run instead of one pair of sleeps per
  scenario.
- `lambda-layer-perf-test/lambda-lite-sdk-run.sh` — standalone lite SDK perf run.

### Terraform
- `terraform/python/lambda-lite-sdk/` and `terraform/node/lambda-lite-sdk/` —
  Lambda function + API Gateway proxy for each runtime.

### Validator
- Registered `PYTHON_LAMBDA_LITE_SDK_INVOKE_TRACE` and
  `NODE_LAMBDA_LITE_SDK_INVOKE_TRACE` templates in `PredefinedExpectedTemplate.java`.
- Added trace-validation configs and expected-data mustache templates for both runtimes.

## Testing
- `./gradlew :validator:build` passes (incl. the test that verifies every
  registered template enum resolves to a real file).
- `lambda-perf-test-all.sh` smoke-tested live against a deployed Python function
  (`NUM_TEST_RUNS=2`); all three scenarios produced valid `Average_Init_Duration`
  data in the expected ordering (Full SDK > Lite SDK > No Layer).
- [NodeJS Instrumentation Main Build PASS](https://github.com/aws-observability/aws-otel-js-instrumentation/actions/runs/28968338850)
- [Python Instrumentation Main Build PASS](https://github.com/aws-observability/aws-otel-python-instrumentation/actions/runs/28968377461)


terraform/node/lambda-lite-sdk/lambda/main.tf
```diff
@@ -0,0 +1,72 @@
...
+      ]
+      resources = [
+        "*"
+      ]
+    }
+  }
+}
+
+module "api-gateway" {
+  source = "../api-gateway-proxy"
```
ezhang6811: Just point this to the existing proxy (../../lambda/api-gateway-proxy), then you can delete yours entirely.

terraform/node/lambda-lite-sdk/api-gateway-proxy/main.tf
ezhang6811, line 1: I think these api-gateway-proxy tf files are the exact same as the default lambda. Can we remove and just have our lambda setup point at the existing one?

terraform/node/lambda-lite-sdk/lambda/main.tf
ezhang6811, line 1: This is also nearly the same as the existing Lambda terraform definitions, except for the env var and the function name. Consider also moving this into the main folder with some branching logic so that we don't need separate terraform files for the lite sdk at all. But ok to keep if too complex to merge.

validator/src/main/resources/expected-data-template/node/lambda-lite-sdk/lambda-invoke-metric.mustache
ezhang6811, line 1: This template also looks the same as the existing template. Can we just point our validation.yml at the existing one?

validator/src/main/resources/expected-data-template/node/lambda-lite-sdk/lambda-invoke-trace.mustache
ezhang6811, line 1: We've added more complexity here. Can we just replace the original template with this one and have both the default and lite sdk tests point at it, same as above.

ezhang6811 (approved):

PR #656, opened 2026-06-25
Add Node EC2 E2E test for Dynamic Instrumentation

liustve:
## Summary
Followup to the Python (#639) and Java (#652) EC2 DI tests, mirroring their 4-cell closed-loop structure for the **ADOT Node** agent: PUT breakpoint via control plane → agent polls and applies → traffic hits breakpoint → snapshot lands in `/aws/service-events/<service>` → assert snapshot fields → DELETE.

- New terraform variant `terraform/node/ec2/adot-di/` — single-instance, nvm + `npm install` ADOT, `node --require .../register`, port 8000, `OTEL_AWS_DYNAMIC_INSTRUMENTATION_*` env. Modeled on `node/ec2/default` + the Python/Java `adot-di` variants.
- 4 validator templates + validations targeting `frontend-service/index.js`:
  - `lineBreakpoint` → `outgoingHttpCall` (line 86, capturing the `options` local)
  - `methodProbe` → `awsSdkCall`
  - `methodBreakpoint` → `remoteService`
  - `methodProbeException` → `mysql` (uncaught exception)
- Reuses `util/di/di-config-manager.py`; adds `di-api-requests-node.json` (Node `CodeLocation` needs `Language=JavaScript` + `FilePath` + `MethodName`). Driver keys off `DI_REQUESTS_FILE` so one driver serves all three languages.
- `node-ec2-adot-di-test.yml` runs the Node 18/20/22/24 matrix.
- Wired `node` into the shared `actions/di` composite action.

- passing workflow: https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/28210642503/job/83571013217


validator/src/main/resources/expected-data-template/node/ec2/adot-di/di-snapshot-line-breakpoint.mustache
```diff
@@ -0,0 +1,86 @@
...
+          "locals": {
+            "bucketName": {
+              "type": "^string$",
+              "value": "^e2e-test-bucket-name-{{testingId}}$"
+            },
+            "warningMsg": {
+              "type": "^string$",
+              "value": "^This is a custom log for validation testing$"
+            },
+            "req": {
```
srprash: why are we capturing and validating local variables like request, response, and s3 client? if they changed with some new node version in the future, these tests will start to break. I don't see much value in capturing or validating these. If we really want to validate the capture of complex objects, we should update the sample app for such cases which will be entirely in our control.

liustve: sounds good, removed the validation

srprash (approved):

PR #657, opened 2026-06-25
feat: add Lambda lite SDK cold-start performance tests (Python and Node)

wespil:
## Description

Adds Lambda lite SDK cold-start performance coverage to the existing layer performance
workflows. The shared `lambda-layer-run.sh` script gains a third **`lite`** mode (alongside the
existing full-SDK and baseline modes), so the Python and Node layer workflows now measure all
three cold-start scenarios — **Full SDK**, **Lite SDK (fast start)**, and **No Layer (baseline)** —
within a single workflow, without needing dedicated lite-SDK workflows.

## Changes

- **Extended** `lambda-layer-run.sh` with a `lite` run mode (`OTEL_AWS_LAMBDA_FAST_START=true`,
  exporters disabled) that writes `lite_sdk_results.txt`. The existing `false` (Full SDK →
  `layer_results.txt`) and `true` (baseline → `no_layer_results.txt`) modes are unchanged, so
  the dotnet/java workflows that call the script are unaffected.
- **Updated** the Python and Node layer workflows to add a "Lite SDK (Fast Start)" step that
  invokes `lambda-layer-run.sh lite`, and to upload the new `lite_sdk_results.txt` artifact
  alongside `layer_results.txt` and `no_layer_results.txt`.

## Why

Adding a `lite` mode to the existing script keeps all three cold-start scenarios in one pair of
layer workflows (Python + Node) and reuses the same query/parsing logic, rather than introducing
separate lite-SDK workflows that would duplicate the YAML and produce parallel result files.

## Testing

Validated end-to-end against real Lambda deployments for files changed: 

- [Python Lambda Layer Performance Test](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/28397060668)
- [Node Lambda Layer Performance Test](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/28397018003)
- [DotNet Lambda Layer Performance Test](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/28396888241)
- [Java Lambda Layer Performance Test](https://github.com/aws-observability/aws-application-signals-test-framework/actions/runs/28397042541)

All three result files were produced in both runs, with init durations in the expected order
(`no-layer < lite-SDK < full-SDK`), confirming each scenario is correctly attributed to its own
CloudWatch time window.


lambda-layer-perf-test/lambda-layer-run.sh
```diff
@@ -3,12 +3,16 @@
 
 #!/bin/bash
 
-IS_BASE_RUN=${1:-false}
```
liustve, line 6: nit: if you're reusing this variable to define modes for, I would just rename it to something that better reflects the 3 options since it's no longer true/false

.github/workflows/node-lambda-layer-perf-test.yml
```diff
@@ -123,6 +129,7 @@ jobs:
         path: |
             no_layer_results.txt
             layer_results.txt
+            lite_sdk_results.txt
```
liustve: lite_layer_results.txt? to keep names consistent

.github/workflows/python-lambda-layer-perf-test.yml
```diff
@@ -79,6 +79,12 @@ jobs:
         chmod +x lambda-layer-perf-test/lambda-layer-run.sh
         ./lambda-layer-perf-test/lambda-layer-run.sh false aws-opentelemetry-distro-python
 
+    - name: Run Cold Start Iterations for Lite SDK (Fast Start)
```
liustve:
```suggestion
    - name: Run Cold Start Iterations for Base Lambda + Lambda Layer Lite (Fast Start)
```

liustve (approved):

PR #674, opened 2026-07-29
FIX main build IRSA permissions issue with CW Agent

haneric00:
*Issue description:*
Previous changes missed a call to clean-app-signals.sh which deleted IRSA permissions and  due to parallelization of EKS tests and all EKS tests now sharing a cluster, a race condition prevents these permissions to be added back to the cluster, resulting in the next main build failing.

*Description of changes:*

This PR removes all references to clean-app-signal.sh during EKS tests. However this script is necessary in flushing the environment between builds. so I added a separate workflow that runs this script to clean the environment.

successful runs: https://github.com/aws-observability/aws-otel-python-instrumentation/actions/runs/30499340766, https://github.com/aws-observability/aws-otel-python-instrumentation/actions/runs/30499340766


v-iw (approved):
Why we need it: A leftover clean-app-signals.sh call deletes the shared CW-agent IRSA. With EKS tests now parallel on one cluster, a race stops it being re-added and fails the next main build.

Why it works: Drops per-test teardown and moves it into one reusable end-of-pipeline workflow; next run's first EKS job reinstalls via enable-app-signals.sh as before.

Why safe: Teardown only runs when the cluster is idle and gated by the caller's needs: on all EKS jobs, plus an internal guard that skips if any ns-* test namespaces remain.

vastin (approved):

PR #675, opened 2026-07-31
test: tighten EKS OCB Environment and PlatformType validation

liustve:
Removes the remaining backward-compat looseness from the `eks-otlp-ocb` templates now that the backend fully emits the new EKS values.

- **PlatformType** (trace/log): `^(Generic|AWS::EKS)$` → `^AWS::EKS$`
- **Environment** (metric): `ANY_VALUE` → `eks:default`

`CWMetricValidator` has no regex support for dimension values, but a literal needs none. `eks:default` is what the OCB path emits: the collector config enables `k8s.cluster.name` but never populates `k8s.namespace.name`, so the backend takes its `eks:default` fallback. This matches the actual validator output in recent OCB CI runs.

Environment is not used as a CloudWatch ListMetrics filter dimension, so a mismatch fails loudly with `EXPECTED_METRIC_NOT_FOUND` rather than silently returning no metrics.

The trace/log Environment regex keeps its `(default|{{platformInfo}}/{{appNamespace}})` alternation, since those validators support regex and the fallback branch costs nothing there.


liustve:
Superseded, reopening with a cleaner branch.

PR #682, opened 2026-08-07
Add additional parallel testing pipeline for Python EKS tests

haneric00:
EKS tests for ADOT Python are currently designed to run sequentially. In order to cover the versions of Python currently supported, this means dedicating for almost an hour for tesking EKS each time we build ADOT Python. Parallelizing EKS tests could remove a major bottleneck in our build/release pipeline. Instead of modifying the existing tests to run in parallel we add additional workflows that are designed from the ground up to run in parallel. This prevents hidden issues from arising related to changing existing code, as well as allows teams that already depend on the existing EKS tests to continue using them without needing to update their build workflows.

*Description of changes:*

Modify existing terraform files to work with new parallel tests.
 - main.tf
 - variables.tf
 - kubeconfig.tpl

*note: due to the nature of parallel execution, certain metadata in main.tf was changed. This change MAY break sequential execution, in which case users should migrate to use a single call of the parallel tests instead.*

Add testing files to mirror the existing sequential EKS test files
- .github/workflows/python-eks-parallel-test.yml
- .github/workflows/eks-setup-app-signals.yml
- .github/workflows/eks-cleanup-app-signals.yml

eks-setup and eks-cleanup exist to enable parallel execution of tests without launching new clusters. In order to prevent race conditions in the setup and cleanup scripts (when multiple scripts are launched in parallel) we create specific setup and cleanup jobs. 
eks-setup:
- installs app signals on the shared cluster before parallel jobs run
- verify CW agent is active and has IRSA creds

eks-cleanup: 
- reap leftover ns-* namespaces
- run cleanup script
- sweep remnants from test environment: delete amazon-cloudwatch-observability addon explicitly, delete CW agent service account, etc.

python-eks-parallel-test.yml models the existing python-eks-test.yml. Upon reviewing a diff between these two files the major differences are:
 - Per-job isolation plumbing (renames/scoping)
 - logic for "verify shared cluster + patch operator + confirm injection" to replace enable-app-signals.sh call
 - removal of logic that would thrash under parallel execution
 - also includes redundant checks that CW agent is active and has IRSA creds

Don't merge until testing branch pins are removed.


.github/workflows/python-eks-shared-cluster-test.yml
jj22ee, line 1: Since this workflow tests only one Python version, I think it should be called something other than `"parallel"`. The main difference for now seems to just be that it doesn't run setup/cleanup scripts, since the caller workflow that runs this workflow multiple times in parallel handles the setup/cleanup.

terraform/python/eks/main.tf
```diff
@@ -59,6 +59,22 @@ provider "kubectl" {
...
+# cluster without "port is already allocated" collisions (NodePort is cluster-wide, not namespaced).
+# Backward-compatible with the sequential test: it passes no python_version, so the default "3.10"
+# yields offset 0 -> 30100 / 30101, identical to the previously hardcoded values.
+locals {
+  version_offset = {
+    "3.10" = 0
+    "3.11" = 1
+    "3.12" = 2
+    "3.13" = 3
+    "3.14" = 4
```
jj22ee: This now becomes a location in which we have to manage regularly if we want to support/deprecate an Python version. Can we populate this as python-version array (each index represents the offset), where the array value is an input from the calling workflow? The calling workflow is usually where we manage supported Python versions today, but those versions might not be managed in an `array` type variable at the moment.

srprash (approved):

v-iw (approved):

